PR #32, opened 2017-08-10
Fix equalsOneOf

aslakhellesoy:
A few tests were failing on Node v8.2.1, where the name of an Assertion Error isn't `AssertionError`, but `AssertionError [ERR_ASSERTION]`.

This fixes that.


ept:
Nice, thank you very much @aslakhellesoy!

PR #33, opened 2017-08-11
Test illustrating non deterministic string array merge

aslakhellesoy:
I'm experimenting with text (as an array of characters) and discovered that merging of concurrent updates to a string array is non deterministic. The test in this PR runs the same test 1000 times:

```
./node_modules/.bin/mocha -g "merges string arrays"
```

On my machine (Node 8.2.1, OS X), the test usually fails within the first 1-3 iterations. However, if you uncomment the deterministic `uuid` function in `automerge.js`, it deterministically fails on iteration 24. Try initialising `n` to a different value than `0`, and it will deterministically fail after a different number of iterations (!)

### DIFF
--- a/src/automerge.js
+++ b/src/automerge.js
@@ -1,5 +1,7 @@
 const { Map, List, fromJS } = require('immutable')
 const uuid = require('uuid/v4')
+// let n=0
+// const uuid = () => "UUID-" + n++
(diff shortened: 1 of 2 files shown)


aslakhellesoy:
Superseded by #34, which is a more minimal example than this one.

PR #34, opened 2017-08-12
Test illustrating bug in Array insertion

aslakhellesoy:
I believe I've found a bug in Array insertion following a merge. It seems like the `actorId` affects the behaviour - with some values it works as expected, with others it doesn't.

The `lamportCompare` function sorts ops based on the `actorId`, and the result of `insertionsAfter` seems to be missing some ops depending on the alphabetical order of the two actors' `actorId`. When the `actorId` isn't specified, the behaviour is non-deterministic (because the `actorId`s are random uuids).

I haven't been able to find a fix, but hopefully these pointers will help someone more familiar with the algorithm and code find and fix the bug quicker!

Cheers,
Aslak

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -530,21 +530,24 @@ describe('Automerge', () => {
       assert.deepEqual(s2._conflicts, {})
     })
 
-    it('should handle insertion', () => {
-      for (let i = 1; i < 10; i++) {
-        try {
-          s1 = Automerge.init()
-          s2 = Automerge.init()
+    it('should handle insertion AB', () => {
+      s1 = Automerge.init('A')
+      s2 = Automerge.init('B')
 
-          s1 = Automerge.changeset(s1, doc => doc.list = ['one', 'three'])
-          s2 = Automerge.merge(s2, s1)
-          s2 = Automerge.changeset(s2, doc => doc.list.splice(1, 0, 'two'))
-          assert.deepEqual(s2.list, ['one', 'two', 'three'])
-        } catch (err) {
-          err.message += `\nFailed on iteration ${i}`
-          throw err
-        }
-      }
+      s1 = Automerge.changeset(s1, doc => doc.list = ['two'])
+      s2 = Automerge.merge(s2, s1)
+      s2 = Automerge.changeset(s2, doc => doc.list.splice(0, 0, 'one'))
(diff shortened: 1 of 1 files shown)


aslakhellesoy:
Upon further investigation, it seems related to the [filter](https://github.com/automerge/automerge/blob/e46ca46fbb796752b665f826db16751f57d6072b/src/op_set.js#L260) in `getNext`. I can make the failing test pass by changing `sib < key` to `sib > key`, but that makes other tests fail.

I'm starting to wonder - am I using the library incorrectly, or is this a bug?

ept:
Hi @aslakhellesoy, thanks for the report. I'm busy right now, but will dig into it tomorrow at the latest.

ept:
@aslakhellesoy This was indeed a bug, and I have fixed it in e368896. Thank you for the nice clear test case!

aslakhellesoy:
Thanks for the quick turnaround @ept. Glad it was easy to fix.

Having worked with ShareJS and a few other OT/CRDT libs, I'm really happy to have found Automerge. I really like how it's decoupled from transport, and the API is just a joy to use.

Keep up the good work!

ept:
Thanks for your kind words. If you have any ideas for improving Automerge, or want to share any more about what you're trying to build, we'd love to hear.

BTW, I've published version 0.4.3 of the automerge npm package that include this bugfix.

Also, I should warn you that right now, the performance for char-by-char text editing is not good, but I'm going to be working on that in the next few weeks.

aslakhellesoy:
I plan to build a few things:

- Collaboration tool for [Example Mapping](https://cucumber.io/blog/2015/12/08/example-mapping-introduction) - similar to trellis (cards), but also with collaborative editing of card text. Webapp, not Electron 
- Collaborative text editor with [Codemirror](https://codemirror.net/)
- Collaborative rich text editor (not sure which one). Maybe [Slate](http://slatejs.org/) or [Quill](https://quilljs.com/) - probably not [Prosemirror](https://prosemirror.net/) - I'd much rather use CRDT than the collab editing protocol/library they've built.

Looking forward to more performant text editing!

ept:
Very cool. I'm also interested in working towards all of those things, so let's stay in touch about work on those integrations. I'll invite you to a Slack channel that we've been using for discussion and planning (feel free to ignore if you don't want to use Slack).

Issue #31, opened 2017-07-04
Document/channel concept


ept:
We have previously discussed a few things we'd like to be able to do with Automerge documents:

* Forking documents (like forking in github/branching in git), with the choice of whether or not to remain subscribed to upstream documents
* Allowing a forked document to be merged back into its upstream document
* Tentatively applying a bunch of changes to a local copy of a document, discarding them again if the result is not as desired, or publishing the merged result if the result is good
* Template documents that set up some basic scaffolding required for an application, and concrete documents that "inherit" from the template
* A mechanism for consuming feeds and merging them into a single document — think RSS feeds, or feeds of currency exchange rates, or suchlike

I had a call with @pvh to discuss how best to implement these concepts in Automerge/MPL. The following are some notes on what we discussed.

As a first step, hash-chaining to encode the dependency graph (#28, like parent commit hashes in git) seems like a good idea. To find all changes that have gone into a document, start with one or more HEAD commit hashes, and traverse the dependency graph. When two communicating nodes have made concurrent changes, they won't know about each others' heads, so they'll need to run a multi-round protocol to figure out their latest common ancestor hash, much like git (#27).

This leaves the question of how you find out about changes to a document. Our proposal is to separate it into two concepts:

1. A *channel* is a network abstraction for pub/sub. A change is published to a channel, and a node can subscribe to any number of channels. A channel has a unique identifier, e.g. a UUID. Channels are probably not visible to the end user, but only an internal abstraction.
2. A *document* is a set of channels, and it incorporates all changes that appear in any of its channels. A document may exist only on one node, and is not necessarily shared with other nodes. To share a document, its set of channels should perhaps be written to a filesystem CRDT?

The features outlined above can all be implemented using those two concepts:

* Forking a document means publishing any future changes to that document to a new channel. The document can remain subscribed to the upstream channels (continuing to incorporate upstream changes), or unsubscribe from the upstream channels.
* Merging a forked document back upstream would mean adding a change with a dependency on the fork to the upstream channel. (I guess that implies that a change's dependencies do not have to be in the same channel, but could be in any channel?)
* Tentatively applying changes could mean adding the experimental channel to the set of channels for a document; as long as this dependency is not published back to the channel, it can be reverted by simply removing the channel again.
* Template documents: if document B inherits from template A, then B includes A's channels as well as its own.
* Likewise, merging feeds into a single document is easily achieved by having each feed in a separate channel, and having the document refer to all of them (so the document is the union of the changes in all of the channels).

[...]

Issue #28, opened 2017-06-22
Reduce size of vector clocks


ept:
At present, we create a new actor ID every time a node is started (either as a completely fresh node, or loading its state from file). That approach ensures that even if you run a process twice on the same machine, they will have different actor IDs, and so they won't step on each others' toes. However, the downside is that the rate of churn of actor IDs is high, and so the vector clocks (which have an entry for each actor) get rather large.

I did a first step in this direction in 9d8f1369d3364674b52090ca4b9923f16f4b8104: instead of including the full vector clock in each changeset, we now only include the actor ID of the originator, the sequence number (which starts at 1 for a new actor ID and is incremented on every changeset), and a minimal set of dependencies (as actorId-seqNum pairs). The set of dependencies does not include any dependencies that can be reached transitively through other dependencies, and it implicitly includes seqNum–1 for the same actorId. In other words, it only references changesets from other actors that were received by the originator between seqNum–1 and seqNum, making the dependencies much like a merge commit in git.

However, the API for figuring out which changesets to send between peers (`getVClock()`, `getDeltasAfter()`) still uses full vector clocks without any truncation. Here, reducing the size is a little trickier. It can be done, but it will require more than one round-trip between peers to figure out what deltas need to be sent.

The reason is as follows: for example, say the latest changeset known by peer A (the "HEAD" in git terminology) has sequence number 42. All other changesets known by A can be reached transitively by following the dependency chains from A:42. So it is technically sufficient for `getVClock()` to return `{A:42}`, since that contains all the necessary information. However, any peer that does not have recent changesets from A cannot interpret the clock `{A:42}`. All it can tell is that it is missing a bunch of changesets from A, but it may also be missing changesets from other actors, and it won't find out what those are until it has received the changesets from A.

To solve this, we can probably take a look at the [git transfer protocol](https://git-scm.com/book/en/v2/Git-Internals-Transfer-Protocols) as implemented by [git-send-pack](https://git-scm.com/docs/git-send-pack) and [git-upload-pack](https://git-scm.com/docs/git-upload-pack). Git has a similar problem, since it identifies the state of a repository by a commit hash, and if you only know a commit hash, you don't know what history you're missing in order to get there.

pvh:
Is this basically done, then, Martin? If so we might close this issue and reopen a new one that described any additional work you envision.

ept:
[...]

PR #36, opened 2017-08-18
Connection abstraction to manage peer communication

ept:
This PR removes the getVClock/getDeltasAfter/applyDeltas API for network communication, and introduces a new Connection API instead. It remains abstract (not tied to any particular network protocol), but it incorporates most of the features [MPL's DeltaRouter](https://github.com/automerge/mpl/blob/f31c343787584baf73aec1c234f9d2db54baf2cd/src/mpl/network/delta-router.js). Hopefully this will let us move complexity out of MPL.

Automerge's new Connection class includes the network protocol essentials that are needed regardless of the physical type of network. The only thing it assumes is that communication between nodes goes via one-to-one (unicast) communication. It supports any network topology, including client/server or peer-to-peer (partially or fully connected graph). It assumes that network connections are slightly stateful (at the level of, say, a TCP connection), but it doesn't assume any long-term state about peers. This connection abstraction is the first step towards using hash chaining instead of vector clocks for tracking causal dependencies (#28).

This PR also explores how to handle multiple documents. It introduces a simple DocSet collection that maps some kind of docId to Automerge documents. I'm not sure yet how this may relate to the *channel* concept discussed in #31. More thinking to be done here.

@pvh if you have time at some point, could you take a look and let me know if you think this is going in the right direction? I'm about to go on holiday for a week (back on Aug 28), and when I get back I can try porting MPL to this Connection API to see how it goes. Also, the tests are currently very verbose; I have in mind a little DSL that would tidy them up, but I haven't yet finished it.


ept:
Updated MPL to use this new Connection API in automerge/mpl#11. Also added the promised DSL to make the tests nicer.

src/connection.js
```diff
@@ -0,0 +1,112 @@
+const { Map, fromJS } = require('immutable')
+const OpSet = require('./op_set')
+
+// Returns true if all components of clock1 are less than or equal to those of clock2.
+// Returns false if the clocks are incomparable or if clock1 is greater than clock2.
```
pvh: When would a clock be incomparable? Reading the code I don't see an obvious case when that would be.

src/connection.js
```diff
@@ -0,0 +1,112 @@
...
+function lessOrEqual(clock1, clock2) {
+  return clock1.keySeq().concat(clock2.keySeq()).reduce(
+    (result, key) => (result && clock1.get(key, 0) <= clock2.get(key, 0)),
+    true)
+}
+
+function mergeClock(clockMap, docId, clock) {
```
pvh: Minor style suggestion - call this operation "union" or "max" to be more explicit about the mechanism.

src/connection.js
```diff
@@ -0,0 +1,112 @@
...
+// Keeps track of the communication with one particular peer. Allows updates for many documents to
+// be multiplexed over a single connection.
+//
+// To integrate a connection with a particular networking stack, two functions are used:
+// * `sendMsg` (callback passed to the constructor, will be called when local state is updated)
+//   takes a message as argument, and sends it out to the remote peer.
+// * `receiveMsg` (method on the connection object) should be called by the network stack when a
+//   message is received from the remote peer.
+//
+// The documents to be synced are managed by a `DocSet`. Whenever a document is changed locally,
```
pvh: Let's talk about docId soon. It's a gnarly hack that I'm concerned will become too ensconced to excise conveniently.

src/connection.js
```diff
@@ -0,0 +1,112 @@
...
+// call `setDoc()` on the docSet. The connection registers a callback on the docSet, and it figures
+// out whenever there are changes that need to be sent to the remote peer.
+//
+// knownClock is the most recent VClock that we think the peer has (either because they've told us
+// that it's their clock, or because it corresponds to a state we have sent to them on this
+// connection). Thus, everything more recent than knownClock should be sent to the peer.
```
pvh: Style - the names on knownClock and advertisedClock feel a little confusing to me since one is theirs and one is ours. Maybe they should be "theirClock" and "ourClock"? Or more explictly "lastHeardClock" and "lastSentClock"? I feel like the optimistic clock updates are likely to confuse someone very badly at some point because of network channel loss.

src/connection.js
```diff
@@ -0,0 +1,112 @@
...
+  }
+
+  close () {
+    this._docSet.unregisterHandler(this._docChangedHandler)
+  }
+
+  sendMsg (docId, clock, changes) {
+    const msg = {docId, clock: clock.toJS()}
+    this._advertisedClock = mergeClock(this._advertisedClock, docId, clock)
+    if (changes) msg.changes = changes.toJS()
```
pvh: I'd probably move the toJS() out of this class; that's a transport concern.

src/connection.js
```diff
@@ -0,0 +1,112 @@
...
+    this._sendMsg(msg)
+  }
+
+  maybeSendChanges (docId) {
+    const doc = this._docSet.getDoc(docId)
+    const clock = doc._state.getIn(['opSet', 'clock'])
+
+    if (this._knownClock.has(docId)) {
```
pvh: We only send peers changes for documents which we know that they know about? Since the peer is about to send you a clock and ask for everything anyway, is this test useful?

test/connection_test.js
```diff
@@ -0,0 +1,309 @@
+const assert = require('assert')
```
pvh: Good set of tests.

src/connection.js
ept: Two clocks are "incomparable" if one of them is greater in some fields, and the other is greater in other fields. That is, the clocks represent two states that have diverged.

src/connection.js
ept: Done in cec835f

src/connection.js
ept: Yes, I was a little concerned about docId too — let me know what medium you'd like to use to talk.

src/connection.js
ept: Ok, renamed in 876e8c1.

What do you mean with optimistic clock updates? Updating `theirClock` when we send a message? I think that's important so that we don't double-send the same message unnecessarily. I am assuming that the channel is TCP-like, i.e. it won't drop random messages in the middle of a session, but just close the connection if messages are dropped and can't be retransmitted.

src/connection.js
ept: `toJS()` converts Immutable.js objects into plain JavaScript objects. I've tried to avoid Immutable.js objects leaking out of the API, so the conversion needs to happen somewhere, and here seemed like the best place imho.

src/connection.js
ept: When we first establish a connection, we don't know whether the peer has any data at all for a given docId. If the peer already has all the data, or most of the data, we want to avoid sending them stuff they already have. So it's necessary to wait until we hear from the peer as to how much, if anything, they already have for some docId.

ept:
Merged this so that we don't have branches hanging around for too long. We can obviously still iterate on the connection abstraction.

src/connection.js
pvh: Ah, I think that language is a bit imprecise. They are comparable, it's just that neither is less than or equal to the other. When I read "incomparable", I was looking for degenerate data or things like that.

src/connection.js
pvh: I haven't got a specific proposal yet but I hate to see it infecting automerge. Can we push it up into MPL? Can we replace docId with a convention where all documents share a namespace but begin under a docId field? (IOW you might find the title for a document at store[docId][title]).

src/connection.js
ept: It's a standard term in mathematics (since vector clocks encode a [partial order](https://en.wikipedia.org/wiki/Partially_ordered_set)). But anyway, I've (hopefully) clarified the comment in aaee2ad.

src/connection.js
ept: Hmm, not sure about that convention — I think it would interact weirdly with the fact that an Automerge document is immutable. The role of DocSet here is to broker a compromise between the immutability of Automerge core, and the fact that we need to process mutations that come in from the network. I thought the cleanest way of handling state changes due to network activity was to make DocSet and Connection mutable, while keeping the Automerge core immutable.

If we put all the documents in a shared namespace, that would probably mean that the top-level namespace must be mutable, while the documents within it would remain immutable. And I think that would be more confusing than helpful.

Alternatively, it might be possible to restructure the API so that every message from the network goes through the same event loop as local user input. That way, the Automerge.Connection abstraction could also be made immutable. But I'm not convinced that would be better… a network connection is essentially a kind of state machine, with state transitions linked to sending and receiving of messages, and I think that kind of pattern is more naturally handled in a mutable way.

PR #37, opened 2017-09-08
Incremental cache maintenance & Automerge state diffing

ept:
This branch introduces two things that might not seem like they should be related, but actually are closely related:

1. Incremental cache maintenance: for example, if a change is made to a list, we make only the minimal modification to the cached list object, rather than rebuilding the whole thing after every changeset. This brings a significant performance improvement: in my test on a 1,000-element list, single-element insertion or deletion becomes about 20x faster (from about 40ms down to 2ms per operation). Still not blazingly fast, but let's just say… it's much less bad now than it was previously.

2. Computing the diff between two Automerge objects, in terms of the edits that you would need to make in order to transform the old state into the new state. This is useful for integrating Automerge with existing editor components, as in aslakhellesoy/automerge-codemirror#1 for example.

This branch still needs some tests, which I'll add next week. I just wanted to share something quickly now since I just got it working. The diff functionality isn't documented yet, but this is what it looks like, in a nutshell:

```js
let state1 = Automerge.changeset(Automerge.init(), doc => doc.text = [])
let state2 = Automerge.changeset(state1, doc => doc.text.push("a"))
let state3 = Automerge.changeset(state2, doc => doc.text.push('b'))

Automerge.diff(state1, state2)
// [ { action: 'insert',
//     objectId: '9ad394d7-8e4a-498f-af47-2e9d1266e2ae',
//     index: 0,
//     value: 'a' } ]

Automerge.diff(state1, state3)
// [ { action: 'insert',
//     objectId: '9ad394d7-8e4a-498f-af47-2e9d1266e2ae',
//     index: 0,
//     value: 'a' },
//   { action: 'insert',
//     objectId: '9ad394d7-8e4a-498f-af47-2e9d1266e2ae',
//     index: 1,
//     value: 'b' } ]
```

I have probably introduced exciting new bugs as well. Let me know if you find any…

### DIFF
--- a/README.md
+++ b/README.md
@@ -161,11 +161,21 @@ state2 = Automerge.changeset(state2, 'Delete card', doc => {
 // Haskell' was set to true, and that 'Rewrite everything in Clojure' was
 // deleted:
 
-state1 = Automerge.merge(state1, state2)
+let finalState = Automerge.merge(state1, state2)
 
 // { cards:
 //    [ { title: 'Rewrite everything in Haskell', done: true },
 //      { title: 'Reticulate splines', done: false } ] }
+
+// If you want to know what changed from one version to another, you can get a
+// diff describing the changes (the objectId is a unique identifier for a
+// particular card object):
+
+Automerge.diff(state2, finalState)
+// [ { action: 'set',
+//     objectId: 'ef794659-d8fa-4e89-8605-06dff020513d',
+//     key: 'done',
+//     value: true } ]
 
 // As our final trick, we can inspect the change history. Automerge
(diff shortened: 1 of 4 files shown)


schrepfler:
Does diff imply locality? Will the network protocol switch mode to some sort of delta/diff mode if so or is auto merge merging continuously?

ept:
@schrepfler This diff function is intended purely for the benefit of a local user interface (e.g. to update the contents of a text editor, or to visually highlight changes). Diffing is not used to sync state over the network — the existing Automerge mechanism is used for that purpose.

PR #41, opened 2017-11-08
Add Automerge.Text datatype

ept:
This is the continuation of #40, improving the performance of text editing (a long list of characters). Before this PR, the slowest part of text editing was the shallow clone of the JS array holding the text that was required on every change (in order to keep it immutable). In this PR, I introduce a new datatype `Automerge.Text`, based on the skip list introduced in #40, which does not need to be cloned in order to be updated.

On a 10,000-operation text editing benchmark, the runtime is reduced to about 10 seconds, which is 7x better than the 73-second runtime before #40. The CPU time per operation is about 1 ms, which should be good enough for most practical purposes. At this point half the time is being spent in GC, and there are no obvious remaining bottlenecks in the code, so any future performance gains will probably have to come from being more careful with heap allocation.

Part of this PR is a significant refactoring that separates the cache maintenance of materialised JS objects from the CRDT internals. The OpSet now doesn't perform any cache maintenance, but instead returns a list of edits that need to be applied to cached objects. The FreezeAPI module applies these edits to the frozen (immutable) JavaScript objects that are returned to the application. This refactoring should also make it easier to support alternative APIs, e.g. an Immutable.js-based API, as suggested on Slack.

The API of `Automerge.Text` is very minimal right now. In the future we'll probably want to expand it to support access patterns that are needed in text editing (e.g. line-based access rather than character-based access; splitting strings into characters). But I want to understand the APIs of existing JS text editors first, and explore how best to integrate with them.

### DIFF
--- a/src/automerge.js
+++ b/src/automerge.js
@@ -3,6 +3,7 @@ const uuid = require('uuid/v4')
 const { rootObjectProxy } = require('./proxies')
 const OpSet = require('./op_set')
 const FreezeAPI = require('./freeze_api')
+const { Text } = require('./text')
 const transit = require('transit-immutable-js')
 
 function isObject(obj) {
@@ -29,7 +30,10 @@ function createNestedObjects(state, value) {
   if (typeof value._objectId === 'string') return [state, value._objectId]
   const objectId = uuid()
 
-  if (Array.isArray(value)) {
+  if (value instanceof Text) {
+    state = makeOp(state, { action: 'makeText', obj: objectId })
+    if (value.length > 0) throw 'assigning non-empty text is not yet supported'
+  } else if (Array.isArray(value)) {
     state = makeOp(state, { action: 'makeList', obj: objectId })
     let elemId = '_head'
     for (let i = 0; i < value.length; i++) {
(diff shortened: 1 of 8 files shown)


pvh:
It's hard to review this patch without any internal documentation. I suppose it would probably be easier to just re-review all of automerge at the moment since a lot of this is restructuring the internals to separate the freezeAPI. Am I correct in understanding that essentially you have separated creating changes from applying them to the materialized view of the document, thus allowing for batch updates and so on?

ept:
I will merge this to avoid having the branch open for too long, and promise that I will add some internal documentation to explain what's going on!

To answer your question, @pvh, yes: I've separated the core CRDT from the materialized view maintenance. The OpSet is the core CRDT, maintaining the set of operations. `OpSet.addChange()` adds an operation to this set, and returns a "diff" object describing the effect of the change (e.g. "insert a new list element at index i of list object o"). When performing small changes, the FreezeAPI module performs incremental view maintenance by interpreting that diff object and applying it to the materialized view of frozen (immutable) JS objects.

For bulk updates (including loading from file), FreezeAPI ignores the diff objects, and instead materializes the frozen JS objects by querying the OpSet. This allows us to directly create the JS objects corresponding to the latest state of the CRDT, without having to go through all of the past states. The `incremental` argument of `FreezeAPI.applyChanges()` chooses between incremental and bulk-update mode.

PR #42, opened 2017-12-10
feat(signing): Add support for signing each change.

mikeal:
In p2p and offline systems it's often necessary to sign each change so that the "actor" noted in the
change can be validated at any time by any participant. Rather than add a bunch of crypto and strong
opinions about *how* to sign I added support for an additional function in the change method that
can be used to return a signature for any change object. This will allow people to implement their
own signing and validation methods without any additional features in automerge.

### DIFF
--- a/src/automerge.js
+++ b/src/automerge.js
@@ -129,12 +129,13 @@ function applyChanges(doc, changes) {
   )
 }
 
-function makeChange(oldState, newState, message) {
+function makeChange(oldState, newState, message, signer) {
   const actor = oldState.get('actorId')
   const seq = oldState.getIn(['opSet', 'clock', actor], 0) + 1
   const deps = oldState.getIn(['opSet', 'deps']).remove(actor)
-  const change = fromJS({actor, seq, deps, message})
+  let change = fromJS({actor, seq, deps, message})
     .set('ops', newState.getIn(['opSet', 'local']))
+  if (signer) change = change.set('signature', signer(change))
   return applyChange(oldState, change)
 }
 
@@ -168,7 +169,11 @@ function parseListIndex(key) {
   return key
 }
 
-function change(doc, message, callback) {
+function change(...args) {
+  let doc = args.filter(arg => typeof arg === 'object')[0]
+  let message = args.filter(arg => typeof arg === 'string')[0]
+  let [callback, signer] = args.filter(arg => typeof arg === 'function')
+
   checkTarget('change', doc)
   if (doc._objectId !== '00000000-0000-0000-0000-000000000000') {
     throw new TypeError('The first argument to Automerge.change must be the document root')
@@ -176,14 +181,11 @@ function change(doc, message, callback) {
   if (doc._change && doc._change.mutable) {
     throw new TypeError('Calls to Automerge.change cannot be nested')
   }
-  if (typeof message === 'function' && callback === undefined) {
-    [message, callback] = [callback, message]
-  }
 
   const oldState = doc._state
   const context = {state: oldState, mutable: true, setField, splice, setListIndex, deleteField}
   callback(rootObjectProxy(context))
-  return makeChange(oldState, context.state, message)
+  return makeChange(oldState, context.state, message, signer)
 }
 
 function assign(target, values) {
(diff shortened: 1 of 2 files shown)


pvh:
Thanks for the PR, @mikeal. This is absolutely a feature we intend to support but I'm not crazy about every caller having to pass in a signature and get that right. 

Do you think there's a way of supporting configurable signing that might have fewer places to make mistakes? I haven't got a particular plan in mind, but perhaps something along the lines of passing a signing function to init() with a signature that looked something like `function(previous_state, change)`?

I had expected in the mid-term we would be opinionated and choose our own signing strategy but perhaps this will be a better overall solution. @ept is probably sleeping right now but he's probably got higher fidelity input.

pvh:
Poking at this a little more, treating the signing function a bit like the actorId as a run-time property of the state seems to be reasonable but a little bit awkward. I'll post a draft PR.

pvh:
Okay, looks like there's some magic in `proxies.js` to look at more closely but there's a baby fussing on my lap so I probably won't get to it this evening.

mikeal:
`init()` does make sense. I considered putting it there but it was less clear to me what side effects there might be by adding it in that code path.

But, in theory, the "actor" you configure should be the public key of the signer (in most cases) and so it makes a lot of sense to pass the signer in at the same time as you configure the actor.

ept:
Hi @mikeal, thanks for the PR!

I agree with @pvh that an additional parameter on every change is not ideal. Passing a function into `init()` is a possibility, but even there it feels a bit wrong to me, since signing is not really an aspect of the core CRDT data structure.

Instinctively I would regard signing as a protocol-level concern, which would mean it would be better placed in Automerge.Connection, for example. Would that work for your use case?

In general, I expect that signing may well interact with other protocol-level concerns, such as serialization, compression, encryption, and so on. (For example, the signature algorithm may require a serialized byte string to sign, not an arbitrary JavaScript object.) Although we don't have to over-engineer all of that right now, I'd like to be mindful of those things. For example, would it make sense to introduce a kind of "middleware" concept at the protocol level, allowing protocol modules to be plugged in?

pvh:
I'm torn. On the one hand we can claim that automerge is just a data structure and users should be free to compose it for their own purposes. This is essentially how things work today. 

Having now got half-way into figuring out automerge-over-dat it's clear to me that the Connection class is the wrong protocol and abstraction. Building on Hypercore (or dat directly) gives us a different set of semantics and a different set of performance trade-offs versus negotiating our own connections and implementing something like our own gossip protocol as `legion` did.

I guess I'd conclude that I'm largely agnostic about the future but a parameter to init() feels like a reasonable and reversible decision.

mikeal:
@pvh I'd like to look at your automerge-over-dat stuff, is it up yet or are still just working on it locally and haven't pushed yet?

mikeal:
Just FYI, I'm implementing an automerge-over-pouchdb over at https://github.com/mikeal/rattan

pvh:
Took a look at rattan, and it's really exciting to see someone taking automerge in a new direction but I can't really see from the code that is there what kind of uses you're aiming to support quite yet. Would love to hear more about it.

For reference, the work I'm doing is to try and build better support for developers who want to make installable P2P apps in Electron which work offline and support real-time sync where the data is "document-like" and by that I mean that anyone with access should be able to make any kind of legal changes (like a Word doc). 

That's not the end goal, particularly, but it's the kind of app I've been working on so that's what I want.

ept:
In #44 I added `Automerge.getChanges()` and `Automerge.applyChanges()`, which you can use to implement a custom network transfer layer for Automerge, including signatures. You can use it something like this:

```js
// On one node
val previousState = state
state = Automerge.change(state, doc => {
  // make arbitrary change to the document
})
val changes = Automerge.getChanges(previousState, state)
for (val change of changes) change.signature = signer(change)
network.broadcast(changes)

// On another node
val changes = network.receive()
for (val change of changes) checkSignature(change)
state = Automerge.applyChanges(state, changes)
```

I think that will do what you need, without having to add signature support into Automerge itself. What do you think?

mikeal:
@ept this is great, and actually much better than I was imagining, it's just going to take a little bit of code to change how I was previously storing things, but the overall impact of doing things this way is much much better.

ept:
@mikeal Glad to hear! I'll close this issue then, but please feel free to get in touch if you run into any issues or want to discuss something.

PR #49, opened 2018-02-05
Replace several empty guids for ROOT_ID constant

ordepdev:
Followed the same approach from `op_set.js` and replaced
all occurrences of `'00000000-0000-0000-0000-000000000000'`
for a `ROOT_ID` constant.


ept:
Thanks for your patch!

PR #50, opened 2018-02-05
 Use better isObject  & prettier on automerge.js

fxfactorial:


pvh:
Thanks so much for sending this over but I'm going to pass on the reformatting commit.

PR #52, opened 2018-02-06
baby array ] came back

YannzizTay:


ept:
Thank you for the fix!

PR #57, opened 2018-02-13
Fix require error in tests

mmmm1998:
Before changes.
default
After changes.
default


ept:
Thank you for the fix! I never noticed this due to macOS' case-insensitive filesystem.

PR #58, opened 2018-02-13
OpSet as immutable.js Record

pboyer:
`npm run test` and `npm run browsertest` both succeeded without regression. 

Any reason you're not using travis or similar? 😇

### DIFF
--- a/src/op_set.js
+++ b/src/op_set.js
@@ -1,4 +1,4 @@
-const { Map, List, Set } = require('immutable')
+const { Map, List, Set, Record } = require('immutable')
 const { SkipList } = require('./skip_list')
 const ROOT_ID = '00000000-0000-0000-0000-000000000000'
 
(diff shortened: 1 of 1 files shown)


ept:
Thanks! Yes, this should work, although I'm not sure there's much of an advantage over the current use of `Map()`. There are quite a lot of structures in Automerge that currently use `Map`, but might arguably be better represented by records. But I'd like to think a bit more before going on a spree to convert all the code to use records.

Regarding CI, yes, it would be great to use Travis or suchlike. I just haven't had time yet to look into how to set it up. If you'd like to help, that would be a welcome contribution! :)

pboyer:
Hi @ept,

I won't try and espouse the benefits as it's been done elsewhere:

https://tonyhb.gitbooks.io/redux-without-profanity/content/using_immutablejs_records.html

pboyer:
Closing for now.

PR #60, opened 2018-02-15
splice in a change callback returns an array of deleted elements

aslakhellesoy:
This makes the proxy API behave like `Array#splice`


ept:
This looks great. Thanks @aslakhellesoy!

PR #61, opened 2018-02-15
Keep a changelog

aslakhellesoy:
I think it would be great to have a semver-based changelog for this project. Here is the start of one (based on #60, assuming that gets merged).


ept:
Good idea! I will fill in the history so far.

PR #64, opened 2018-02-22
Calling applyChanges twice with unresolved dependencies would throw

jimpick:
Here is a test case, and a workaround.


ept:
Fixed the issue. Thanks @jimpick for the nice clear test case!

PR #69, opened 2018-02-25
generate actorId in Automerge.load when unspecified

saranrapjs:
While playing around with `Automerge.load`, I ran into the `actorId` being undefined when not specified explicitly (which is different from the `.init()` behavior). This branch should make the following true for `Automerge.load`/`Automerge.loadImmutable`:

> Each Automerge instance has an actor ID — a UUID that is generated randomly whenever you do Automerge.init() or Automerge.load() (unless you explicitly pass an actor ID into those functions).

### DIFF
--- a/src/automerge.js
+++ b/src/automerge.js
@@ -162,11 +162,11 @@ function assign(target, values) {
 }
 
 function load(string, actorId) {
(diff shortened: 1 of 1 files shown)


ept:
Hi @saranrapjs, thanks for fixing this! I've also added some tests to catch this kind of thing in the future (embarrassingly, we didn't have any tests for saving/loading).

PR #73, opened 2018-03-09
Fixes npm test on case sensitive file systems

solsson:


fueangfa:
Good.

ept:
Thanks! I have now set up Travis CI to detect this kind of issue in the future.

Issue #71, opened 2018-03-02
Get path of modified objects in diff?


aslakhellesoy:
When an Automerge document is changed as the result of applying changes from a peer, a UI usually needs to be updated. There are two strategies for updating the UI:

* Rerender the UI
* Modify the UI that is affected by the changes without rerendering

In the context of web applications using a virtual DOM (such as React), rerendering often works well. The virtual DOM's diffing algorithm can detect what won't change (most of the virtual dom), and will only rerender a subset of the UI.

There are however situations where this approach doesn't work well. 

One situation is a UI with several independent text editors that can be collaboratively edited, and also 
moved around. For example a card wall application where cards can be dragged around and also edited concurrently.

Consider the case where user A is editing the text in Card X while user B is moving Card X to a different location of the board.

The rerender with virtual DOM approach would detect that the X card has moved, and rerender it. As it's being re-rendered, the text editor component in the old location is removed from the DOM and a new one is created in the new location.

This causes user A to lose focus in their editor, which is not a good user experience.

A better approach would be to *move* the DOM node with the editor component to the new location, allowing user A to keep typing without losing focus while card X is being moved. For React applications this can be achieved with the technique described [here](https://gist.github.com/leoasis/e1d093141e5f22e4e1e346e6726dfa5b) (thanks for the [pointer](https://automerge.slack.com/archives/C61RJCM9S/p1519972025000098?thread_ts=1519662292.000353&cid=C61RJCM9S) @flash1293), but if the application doesn't use React this poses a new challenge.

The application developer now needs to know what card was moved in the Automerge doc, so they can move the corresponding DOM node. This can be determined by examining the diff returned by `Automerge.diff`. The flat nature of this format makes this difficult. If the object representing the moved card is located deep inside the document it can be both difficult and slow to discover the old and new path. One would have to traverse the old Automerge doc and match `_objectId` properties with those in the diff.

Would it be possible to make Automerge return a diff that has more information about the *path* of the modified objects in a document? A path could be represented as an array of property names and indexes, as described [here](https://github.com/ottypes/json0#operations).

ept:
Hi @aslakhellesoy, sorry for the slow response. This seems like a good idea. I've put an initial implementation here: a9d75d9e5c8d0b8ed017573a3babc8807173b311 — could you give it a try and see if it does what you want?

[...]

PR #74, opened 2018-03-12
Include path from root to modified object in Automerge.diff()

mariusdj15:
As suggested in #71

### DIFF
--- a/src/op_set.js
+++ b/src/op_set.js
@@ -36,6 +36,29 @@ function transitiveDeps(opSet, baseDeps) {
   }, Map())
 }
 
+// Returns the path from the root object to the given objectId, as an array of string keys
+// (for ancestor maps) and integer indexes (for ancestor lists). If there are several paths
+// to the same object, returns one of the paths arbitrarily. If the object is not reachable
+// from the root, returns null.
+function getPath(opSet, objectId) {
+  let path = []
+  while (objectId !== ROOT_ID) {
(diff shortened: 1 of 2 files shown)


ept:
There are some caveats about this PR, which is why I've not yet merged it. (The following text is copied from #71.)

At present, Automerge allows an object to appear in several places in the tree (i.e. it allows the document to be a DAG), in which case the path will not be unique. In the future I plan to change this (i.e. I want to restrict the structure to always be a tree). With this in mind, I haven't added support for multiple paths; instead, if there are multiple paths to the same object, the current code just picks one arbitrarily. That behaviour is a bit nasty at present, but will make more sense when the tree restriction is in place.

Another caveat: sometimes an object is created and modified before it is linked into the tree. For example, if you do

```js
Automerge.change(state, doc => doc.foo = {bar: 'baz'})
```

then Automerge first creates an empty map object, assigns the key `bar` on that object to the value `baz`, and then links that new object into the root object under key `foo`. This means that at the time of assignment of the key `bar`, the object is not yet in the tree, and thus the assignment operation of `bar` has no path. If this is a problem, we could probably switch the order of operations so that the empty object is linked into the tree first, and then its key is assigned.

ept:
I have merged this PR, despite the caveats mentioned above, because we need it for integrating text editors with Automerge, and the caveats currently don't seem to be a problem for this use. We can always revisit it again in the future.

Issue #19, opened 2017-05-19
assignment merges


choxi:
Adam found a bug where assignments were getting clobbered on merges, I figured out that it was happening because on assignment updates I was resetting the entire `assigned` object instead of the specific prop:

```javascript
  updateAssignments(state, action) {
    let cardIndex = this._findIndex(state.cards, (c) => c.id === action.cardId)
    return Tesseract.set(state.cards[cardIndex], "assigned", action.assigned)
  }
```

So I tried:

```javascript
  updateAssignments(state, action) {
    let cardIndex = this._findIndex(state.cards, (c) => c.id === action.cardId)

    // cards don't have an "assigned" prop by default,
    // we have to set it if it doesn't already exist
    if(!state.cards[cardIndex].assigned)
      state = Tesseract.set(state.cards[cardIndex], "assigned", {})

    // make the update more granular so we don't clobber the entire "assigned" object
    return Tesseract.set(state.cards[cardIndex].assigned, action.person, action.isAssigned)
  }
```

However, this doesn't work either because now we're clobbering the "assigned" object when we initialize it in the `if` expression. I finally got it to work by making sure every card has the "assigned" prop when it's created (in the seed data or when a user creates a card). You can see the full diff [here](https://github.com/inkandswitch/trellis/commit/8a00afcf5b7d5ed9bbc602f9b0a7cc5ae906c87d).

This works in our use case because no one's using the app so it's easy enough to change how we initialize the data, but if we were using Trellis in production I'd have to write a migration to add an empty "assigned" property to every card and make sure all future cards have it by default. I'm not sure if that would even work because the migration itself could end up clobbering some users' assignment props. Is there a better way to handle this that I haven't thought of?

ept:
Closing this as the issue is discussed in detail in #4.

PR #83, opened 2018-04-14
Make _objectId non-enumerable.

EthanRBrown:
This modifies both the state object and the proxy to make the `_objectId` property non-enumerable, making Automerge more developer-friendly by obscuring implementation details.

Significant updates were required to the tests, as `_objectId` was included in many assertions.  This change has the added benefit of simplifying the tests.

A test was added to cover the use of `Object.assign` (or, equivalently, the proposed object spread operator).  No additional test was necessary to verify that `_objectId` was enumerable because of the aforementioned assertions.

For more details on the motivation for this change, please the following [justification](https://gist.github.com/EthanRBrown/eb7ecc51b74170a5d415a742a6c3ca2b).

### DIFF
--- a/src/freeze_api.js
+++ b/src/freeze_api.js
@@ -8,12 +8,15 @@ function isObject(obj) {
 
 function updateMapObject(opSet, edit) {
   if (edit.action === 'create') {
-    return opSet.setIn(['cache', edit.obj], Object.freeze({_objectId: edit.obj}))
+    return opSet.setIn(['cache', edit.obj],
+      Object.freeze(Object.create(Object.prototype, { _objectId: { value: edit.obj } }))
+    )
   }
 
   const oldObject = opSet.getIn(['cache', edit.obj])
   const conflicts = Object.assign({}, oldObject._conflicts)
   const object = Object.assign(Object.create({_conflicts: conflicts}), oldObject)
+  Object.defineProperty(object, '_objectId', { value: oldObject._objectId })
 
   if (edit.action === 'set') {
     object[edit.key] = edit.link ? opSet.getIn(['cache', edit.value]) : edit.value
@@ -40,9 +43,9 @@ function parentMapObject(opSet, ref) {
   const oldObject = opSet.getIn(['cache', ref.get('obj')])
   const conflicts = Object.assign({}, oldObject._conflicts)
   const object = Object.assign(Object.create({_conflicts: conflicts}), oldObject)
+  Object.defineProperty(object, '_objectId', { value: oldObject._objectId })
   const value = opSet.getIn(['cache', ref.get('value')])
   let changed = false
-
   if (isObject(object[ref.get('key')]) && object[ref.get('key')]._objectId === ref.get('value')) {
     object[ref.get('key')] = value
     changed = true
@@ -192,7 +195,7 @@ function instantiateImmutable(opSet, objectId) {
     obj = Object.create({_conflicts: Object.freeze(conflicts.toJS())})
 
     for (let field of OpSet.getObjectFields(opSet, objectId)) {
-      obj[field] = OpSet.getObjectField(opSet, objectId, field, this)
+      Object.defineProperty(obj, field, { value: OpSet.getObjectField(opSet, objectId, field, this), enumerable: !field.startsWith('_') })
     }
   } else if (objType === 'makeList') {
     obj = [...OpSet.listIterator(opSet, objectId, 'values', this)]
@@ -242,10 +245,11 @@ function applyChanges(root, changes, incremental) {
     newRoot = opSet.getIn(['cache', OpSet.ROOT_ID])
     if (newRoot === root) {
       newRoot = Object.assign(Object.create({_conflicts: root._conflicts}), root)
+      Object.defineProperty(newRoot, '_objectId', { value: root._objectId })
       opSet = opSet.setIn(['cache', OpSet.ROOT_ID], newRoot)
     }
   } else {
-    [opSet, newRoot] = materialize(opSet)
+    ;[opSet, newRoot] = materialize(opSet)
   }
   return rootObject(root._state.set('opSet', opSet), newRoot)
 }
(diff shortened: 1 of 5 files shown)


ept:
Hi @EthanRBrown, thanks for your PR! I've actually gone back and forth on whether `_objectId` should be enumerable or not. You make 

doc.widgetsById[id] = { ...doc.widgetsById[id], cost: 2.25 }

ept:
Hi @EthanRBrown, thanks for your PR! I've actually gone back and forth on whether `_objectId` should be enumerable or not. You make [compelling arguments](https://gist.github.com/EthanRBrown/eb7ecc51b74170a5d415a742a6c3ca2b), but there is still an issue for which I don't have a good answer, and maybe you could provide your opinion.

Consider Problem 2 in your gist, about the pattern for merging objects, or making a copy of an object so that changes can be applied to the copy:

```js
doc.widgetsById[id] = { ...doc.widgetsById[id], cost: 2.25 }

**Comment by ept** [comment]
Hi @EthanRBrown, thanks for your PR! I've actually gone back and forth on whether `_objectId` should be enumerable or not. You make [compelling arguments](https://gist.github.com/EthanRBrown/eb7ecc51b74170a5d415a742a6c3ca2b), but there is still an issue for which I don't have a good answer, and maybe you could provide your opinion.

Consider Problem 2 in your gist, about the pattern for merging objects, or making a copy of an object so that changes can be applied to the copy:

```js
// update the cost of doc.widgetsById[id] using the object spread operator:
doc.widgetsById[id] = { ...doc.widgetsById[id], cost: 2.25 }

// equivalently, using Object.assign:
let widget = Object.assign({}, doc.widgetsById[id])
widget.cost = 2.25
doc.widgetsById[id] = widget
```

`_objectId` is included in the copied object only if the property is enumerable. Now Automerge handles the assignment of `doc.widgetsById[id] = someObject` differently, depending on whether `someObject` has an `_objectId` property or not. If there is no such property, Automerge treats the assignment as the creation of a brand new object; if `_objectId` does exist, Automerge can detect the assignment as an update of an existing object. (There might be bugs, but that is at least the intention.)

The distinction between creating a brand new object and updating an existing object is important when other users are concurrently making changes. When a brand new object is created, and another user concurrently updates the existing object, the concurrent update is discarded, so the other user's change is lost. For example, issue #19 is an instance of this problem, and a related issue is discussed in #4.

To make Automerge combine the concurrent changes correctly requires doing the assignment at the individual field level, rather than replacing the entire object:

```js
// best way of doing the assignment in Automerge
doc.widgetsById[id].cost = 2.25

// another user can concurrently update another property of the object,
// and the changes will be merged:
doc.widgetsById[id].name = 'new name'
```

Since Automerge translates those seemingly-mutating assignments into immutable data structure updates, it's fine to use them in a React/Redux context. So I would encourage developers to use those fine-grained assignments of individual properties when possible, instead of the more verbose and coarse-grained pattern of `doc.widgetsById[id] = { ...doc.widgetsById[id], cost: 2.25 }`.

However, I totally understand that people will nevertheless use the object-copying approach that is common in current React/Redux code. So the question is how best to handle the situation when that happens. Your proposed patch means that if someone uses this pattern, it will be interpreted as the creation of a new object, since Automerge cannot keep track of the fact that it's actually only an update of a property within that object. I think that would not be ideal, since it will work fine in the single-user case, but fail to merge as expected in the multi-user case.

One option that I considered: when an object is assigned, Automerge does a diff between the old and new object, and if there is "some similarity" between them, the assignment is treated as applying the diff, rather than replacing the old object with a brand new object. That would allow more fine-grained merging, but I am concerned that it would be too arbitrary and "magical". For example, how similar would the objects have to be before we switch into diffing mode? The threshold would be totally arbitrary.

I am not sure how to balance these competing interests. @pvh do you have any thoughts?

ept:
One more thought. Problem 3 you highlight is that the `_objectId` properties are of no practical use to users… yet the objects in your example do in fact have a UUID (the keys of the `widgetsById` map). I assume that UUID is generated by the application. But I think it would actually make sense for applications to use the `_objectId` generated by Automerge for precisely this purpose. In that case, the `_objectId` is no longer noise, because it's actually relevant to the application.

j-f1:
> For example, how similar would the objects have to be before we switch into diffing mode?

You could do it only if no properties have been removed in the change.

ept:
Thank you for your thoughtful arguments @EthanRBrown. You've convinced me that it's better to make `_objectId` non-enumerable, and that we need better docs to explain how changes should be structured so that the conflict resolution works. Another reason for making it non-enumerable: on arrays, `_objectId` is already a non-enumerable property (since otherwise array iteration would be broken), and thus making it non-enumerable on objects makes the two consistent.

Your first example (`"doesn't do what you expect"`) is a bug in Automerge: my intention was that the changes to the `bar` and `fizz` should take effect, but clearly it's not working as intended. I'll look into that separately.

Before merging I'm going to make a few changes to the patch for internal code structure purposes (in particular, taking `_objectId` handling out of OpSet) but they should be functionally neutral. Hope to have this done today.

src/freeze_api.js
```diff
@@ -242,10 +245,11 @@ function applyChanges(root, changes, incremental) {
     newRoot = opSet.getIn(['cache', OpSet.ROOT_ID])
     if (newRoot === root) {
       newRoot = Object.assign(Object.create({_conflicts: root._conflicts}), root)
+      Object.defineProperty(newRoot, '_objectId', { value: root._objectId })
       opSet = opSet.setIn(['cache', OpSet.ROOT_ID], newRoot)
     }
   } else {
-    [opSet, newRoot] = materialize(opSet)
+    ;[opSet, newRoot] = materialize(opSet)
```
ept: @EthanRBrown Why is this semicolon necessary? I'm aware of the rule that lines starting with a parenthesis need a leading semicolon, but I didn't know there was such a rule for square brackets.

j-f1: If you were to write this:

```js
foo()
[bar, baz] = quux()
```

It would be interpreted as:

```js
foo()[(bar, baz)] = quux()
```

ept: Got it, thanks!

PR #86, opened 2018-04-20
Minor Babel tooling change.

EthanRBrown:
Switched from babel-polyfill to idempotent-babel-polyfill, a thin
wrapper that simply checks to see if there's a polyfill instance already
before adding it, preventing frustrating collisions when using two
packages which both depend on Automerge.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
   "author": "",
   "license": "MIT",
   "dependencies": {
-    "babel-polyfill": "^6.26.0",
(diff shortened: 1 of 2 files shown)


EthanRBrown:
I am closing this PR; I am learning (the hard way) that `babel-polyfill` shouldn't be used at all for libraries for this exact reason, and instead `transform-runtime` should be used, but I'm having trouble making that work.  Please forgive the PR noise...learning something new every day.

PR #87, opened 2018-04-21
Replaced generators.

EthanRBrown:
Refactored generators to regular functions to eliminate the necessity of
a generator polyfill, which causes a lot of unfortunate distribution
overhead (i.e. the use of Babel's transform-runtime, which I have been
unable to integrate successfuly).  It also reduces the size of the
generated bundle by ~260k (over the previous use of babel-polyfill,
which I have learned is a poor choice to use in libraries).

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,6 @@
   "author": "",
   "license": "MIT",
   "dependencies": {
-    "babel-polyfill": "^6.26.0",
     "immutable": "^3.8.2",
     "transit-immutable-js": "^0.7.0",
     "transit-js": "^0.8.861",
--- a/src/skip_list.js
+++ b/src/skip_list.js
@@ -4,12 +4,17 @@ const { Map } = require('immutable')
 // That is, returns k with probability p * (1 - p)^(k - 1).
 // For example, returns 1 with probability 3/4, returns 2 with probability 3/16,
 // returns 3 with probability 3/64, and so on.
-function* randomLevel() {
-  while (true) {
-    const rand = Math.floor(Math.random() * 4294967296)
-    let level = 1
-    while (rand < 1 << (32 - 2 * level) && level < 16) level += 1
-    yield level
+function randomLevel() {
+  // NOTE: this function used to be a generator; it has been converted to a regular
+  // function (that mimics the interface of a generator) to avoid having to include
+  // generator polyfills in the distribution build.
+  return {
+    next() {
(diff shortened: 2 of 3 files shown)


gedw99:
I support this PR.
Make this easier for golang and dart packages I am working on.

ept:
Thanks @EthanRBrown, I don't mind replacing generators if that helps consumers of the code. I find a bit puzzling that defining generators requires such an expensive polyfill, but iterating over a generator (e.g. `[...OpSet.listIterator()]`, as used in [freeze_api.js:209](https://github.com/automerge/automerge/blob/6f204783a3bd54a840798b982af7bb6495e01f2d/src/freeze_api.js#L209)) apparently doesn't. But if this works, I'm happy!

Do you know if there's some automated way of testing whether the `dist/automerge.js` contains all the necessary polyfills? I think it would be nice to have some test that fails if, for example, someone introduces a generator function.

ept:
I have published version 0.7.7 which contains this PR.

EthanRBrown:
@ept I'm not going to claim to understand all the nuances: I'd have to do a lot more research on that.  To your question, however, iterators are not the same as generators.  That is, `[...OpSet.listIterator]` isn't implemented with a generator, but it does provide an iterator interface (otherwise the array spread wouldn't work).  Iterators are simply a well-defined interface (that doesn't require a polyfill).  It just so happens that generators implement that interface.  (As MDN describes it, generator functions are "iterator factories".)

As for your second question about how to test: the most straightforward way would be simply to point the existing tests to `../dist/automerge` instead of `../src/automerge`.  That would certainly catch the inadvertent use of generators (the tests will fail spectacularly), and I suspect most other polyfill-related issues.

Of course, making that change could add an undue burden on testing in development, which is no good, so I would probably recommend paramaterizing that choice.  Something like this:

```javascript
const Automerge = process.env.TEST_DIST==='1'
  ? require('../dist/automerge')
  : require('../src/automerge')
```
Then you would just set `TEST_DIST=1` in the CI environment.  The problem is that it might cause developer consternation.  Someone makes a change, all the tests pass, and then they fail CI because the tests are running off of transcompiled code.  Everything's a trade-off, isn't it?

Incidentally, if you want to see how things get mucked up with generators, with the latest code (0.7.7+), just add a generator to a source file (like `function* test() { yield 'whatever' }`), run `npm run build` and try running the tests or a demo file.  You'll see it blow up.

PR #90, opened 2018-05-11
Support Node 10

aslakhellesoy:
`yarn install` was failing on Node 10.0.0 because of the transitive dependency `karma` -> `chiokidar` -> `upath 1.0.4`, which has `engine` set to `<9`. This constraint is removed in `upath 1.0.5`, which is why I added that dev dependency. I also upgraded all other libs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ jobs:
       addons:
         sauce_connect: true
     - stage: deploy
-      node_js: 9
+      node_js: 10
       script: yarn install
       deploy:
         provider: npm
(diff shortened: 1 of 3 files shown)


aslakhellesoy:
That seems to have broken the build - not sure why

aslakhellesoy:
I reverted the package upgrades, and the build now passes with node 8,9,10

aslakhellesoy:
Ping :-)

ept:
Sorry for the delay. I've merged this, and also upgraded other dependencies (except for webpack, for which the latest version doesn't seem to be working at the moment — will figure that out separately). Thanks @aslakhellesoy!

PR #91, opened 2018-05-15
Improve performance of bulk assignments

ept:
This PR improves the performance of changes that affect a large number of elements (map keys or list elements) of a single object within a single change block. For example, assigning a 10,000-element list:

```js
let x = [];
for (let i = 0; i < 10000; i++) x.push(i);
let doc = Automerge.change(Automerge.init(), d => d.x = x)
```

Before this PR, the code above takes about 30 seconds to run; after this patch, it takes about 3 seconds. This improvement is by replacing two accidentally-quadratic algorithms with linear-time ones.

### DIFF
--- a/src/auto_api.js
+++ b/src/auto_api.js
@@ -1,4 +1,4 @@
-const { List, fromJS } = require('immutable')
+const { Map, List, fromJS } = require('immutable')
 const OpSet = require('./op_set')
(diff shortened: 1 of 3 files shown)


pvh:
Looks good here. If you cut a release perhaps we can pull it into pushpin.

PR #100, opened 2018-06-18
Follow JS delete behavior

### LINKED ISSUE #97: `delete` operator inconsistent with JavaScript standards
Inside an Automerge `change` callback, the `delete` operator will throw an error if the property being deleted doesn't exist:

```javascript
doc = Automerge.change(doc, '', doc => {
  delete doc.container.nonExistentProp    // throws error
})
```

This is inconsistent with JavaScript's `delete` operator which simply returns `true` if the property was successfully deleted OR it didn't exist in the first place (it only returns `false` if the property exists and couldn't be deleted, such as from a frozen object -- in any rate, it never throws a runtime error).

I believe the API behavior of Automerge should be consistent with JavaScript whenever feasible, and unless there's a good reason for this behavior, I recommend it be changed (would be happy to make that fix myself).

salzhrani:
should fix #97 
sorry for jumping in @EthanRBrown, I am trying to get familiar with automerge

### DIFF
--- a/src/automerge.js
+++ b/src/automerge.js
@@ -105,7 +105,7 @@ function deleteField(state, objectId, key) {
     return splice(state, objectId, parseListIndex(key), 1, [])
   }
   if (!state.hasIn(['opSet', 'byObject', objectId, key])) {
-    throw new RangeError('Field name does not exist: ' + key)
+    return state;
   }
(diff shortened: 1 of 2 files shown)


ept:
This looks great, thank you @salzhrani and @EthanRBrown!

PR #108, opened 2018-07-22
Expose getter for specific objects

### LINKED ISSUE #106: Expose opset cache getter method in proxy
For deeply nested objects,  it's difficult to maintain the path to get a particular object, especially as parents get spliced.
Instead, I propose a method that allows one to get an object by it's `_objectId`.

Current: `doc.foo.bar[5].baz[3].insertAt(0, 'val')`
Proposed: `doc._get(baz3UUID).insertAt(0, 'val')`

For me, it'd be much easier to keep track of an immutable ID vs. a mutable path to the object.

Thanks for automerge!

mattkrick:
fix #106.
This is useful in cases where I know which object I'd like to mutate, but I'm not sure where that object lives in the tree.


ept:
Hi @mattkrick, this looks like a good addition to the API, thank you! I've merged it.

PR #119, opened 2018-09-20
Attach a sequential ID to patches

ept:
The frontend checks that patches are applied in strictly incrementing order. This prevents network implementations from accidentally applying patches in the wrong order, or skipping or duplicating patches.

Since `getPatch()` now requires a state update (namely incrementing the patch counter), I changed the API: instead of `getPatch()` there is now an integrated `load()` function that constructs a new backend document from a list of changes.

### DIFF
--- a/backend/index.js
+++ b/backend/index.js
@@ -129,7 +129,24 @@ function init(actorId) {
     throw new TypeError('init() requires an actorId')
   }
   const opSet = OpSet.init()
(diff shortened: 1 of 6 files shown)


ept:
Decided that we don't want this feature after all; the `load()` approach doesn't work well in Hypermerge, where we build a backend one hypercore at a time. So I'm going to close this without merging.

PR #124, opened 2018-09-25
Use explicit exponentiation instead of literal int in randomLevel()

wincent:
This replaces the magic literal int (4294967296) with explicit exponentiation (`2**32`), which should be a little more intention-revealing.

Babel will transpile this to `Math.pow(2, 32)`, but in all cases perf is basically indistinguishable from the integer constant:

https://jsperf.com/pow-vs-exp-vs-raw/1

### DIFF
--- a/backend/skip_list.js
+++ b/backend/skip_list.js
@@ -10,7 +10,7 @@ function randomLevel() {
   // generator polyfills in the distribution build.
   return {
     next() {
-      const rand = Math.floor(Math.random() * 4294967296)
+      const rand = Math.floor(Math.random() * 2**32)
       let level = 1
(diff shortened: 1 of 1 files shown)


j-f1:
That’s not the result I get on iOS Safari:

Math.pow is 98% slower

Would it be worth it to put the raw number back and put `/* 2**32 */` after it so people can see what it is?

wincent:
That, or assign it to a variable outside the function. Thing is, I am not sure this is in a critical section.

ept:
How about writing the number as a hexadecimal literal? That makes the meaning clearer without any performance impact. d9011a478e504bad456bd6b42317afcd33513b0e

wincent:
Sounds good.

PR #125, opened 2018-09-25
Throw error objects instead of strings in backend/op_set.js

wincent:
Elsewhere in the automerge we consistently throw `Error` objects rather than plain strings, so let's do the same throughout backend/op_set.js as well.


ept:
Thanks!

PR #128, opened 2018-10-18
Fix Text update when mixed with other object updates

CGNonofr:
The `updateTextObject` function is called with a `startIndex` and an `endIndex` parameter.

All diffs between these indexes should refer to the same object.

However we use the first diff (Is there a reason?) which can refer to a completely different object, thus the function will crash


ept:
Thank you, good find! I also added a test that reproduces this issue: 1fcc0c45a7e41fd3c83c8f6492dfee97590477ce

Issue #138, opened 2018-12-06
Cannot use a Date value


m-abs:
Hi,

So far `automerge` looks very promising, but I just ran into this minor problem. 
When I set a `Date` object on my `doc` an empty object is added instead of the `Date`.

```Javascript
 let doc1 = Automerge.change(Automerge.init(), "Initial value", doc => {
    doc.started = true;
    const updated = new Date();
    const updatedJSON = JSON.stringify(updated);
    doc.updated = updatedJSON;
    doc.date = updated;
  });

  console.log(doc1);
  // { started: true, updated: '"2018-12-06T10:18:09.572Z"', date: {} }
```

As you can see, the stringified `date` is stored as expected, but the `Date` object isn't.

Versions:
- automerge@0.9.2
- nodejs@10.12.0

ept:
Ah, that's interesting. Thanks for the report, @m-abs. This is happening because `typeof (new Date()) == 'object'` and `Object.keys(new Date()) == []`, so Automerge is treating the value as the empty object `{}`. Looks like we will need to special-case values where `value instanceof Date`.

What would you expect to happen in this case? Should Automerge turn the Date object into an ISO8601 string, like JSON.stringify, or should it treat Date as a separate datatype that is turned back into a Date object when the document is loaded?

m-abs:
Hi @ept 

When I ran into this I expected it to be an instance of Date :)

As far as I know a `Date`-object can't be frozen, but you might be able to wrap it in a `Proxy`-object and get the same effect.

ept:
On #139 I've implemented the simpler option, which is to store and return the value as a string. What do others think — is this a good API behaviour? @pvh?

If we want to turn it back into a Date object we would need to add some metadata to the operation format that tells us how to interpret a value. That is of course possible. One thing to consider in this context is that we'll want to support Automerge in languages other than JavaScript in the future, so any datatypes we add should be widely supported; however, I imagine all mainstream languages will have a built-in datetime/timestamp datatype, so this shouldn't be a problem.

32bitkid:
@ept Personally, I like the solution in #139, given that Automerge is self-described as JSON-like and there is no JSON date type. The thing that I don't like about it is that it's asymmetric, so if I have to manually handle taking a value _out_, then I almost want to do it manually on both directions rather than relying on an auto conversion one way, but not the other.

[...]

PR #139, opened 2018-12-06
Support Date objects in Automerge documents

ept:
To address #138, this patch detects when a JavaScript Date() object is assigned to an Automerge document, and converts it to an [ISO 8601 string](https://en.wikipedia.org/wiki/ISO_8601). This behaviour matches that of `JSON.stringify()`.

### DIFF
--- a/backend/index.js
+++ b/backend/index.js
@@ -9,16 +9,16 @@ class MaterializationContext {
   }
 
   /**
-   * Unpacks `value`: if it's an object of the form `{objectId: '...'}`, updates
-   * `diff` to link to that objectId. Otherwise uses `value` as a primitive.
+   * Updates `diff` with the properties from `data`. The object `data` should be
+   * of the form `{value: objectId, link: true}` if the value is a reference to
+   * another object, of the form `{value: timestamp, datatype: 'timestamp'}` if
+   * the value is a Date object, and of the form `{value: primitiveValue}`
+   * if the value is a primitive value.
    */
-  unpackValue(parentId, diff, value) {
-    if (isObject(value)) {
-      diff.value = value.objectId
-      diff.link = true
-      this.children[parentId].push(value.objectId)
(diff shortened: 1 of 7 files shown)


ept:
After private feedback I changed this PR: instead of turning Date objects into ISO8601 strings, Automerge now supports Date objects as values anywhere in an Automerge document. Internally, the Date object is converted into a timestamp (consisting of the number of milliseconds since the 1970 epoch, which is the underlying timestamp format of both JavaScript and Java), with a piece of metadata indicating that the timestamp should be interpreted as a datetime object, not a number. Even as we port Automerge to other programming languages, it should be easy enough to support an equivalent timestamp datatype.

ept:
Round-trip example of a Date object being converted to JSON and back again:

```js
let s1 = Automerge.change(Automerge.init(), doc => doc.now = new Date())
let changes = JSON.stringify(Automerge.getChanges(Automerge.init(), s1))
let s2 = Automerge.applyChanges(Automerge.init(), JSON.parse(changes))
s2.now instanceof Date // true
```

j-f1:
Does this support time zones?

ept:
JavaScript Date objects do not store the time zone explicitly; the timestamp is in UTC. However, the Date object provides methods for accessing the date and time both in UTC (`getUTCHours()`, `getUTCMinutes()`, …) and in the user's local time zone (`getHours()`, `getMinutes()`, …).

32bitkid:
I was just going to ask if the persisted time is epoch milliseconds from UTC epoch or local epoch, and happy that it is UTC. RE: timezone-ish metadata. I guess the closest that you could do would be to also persist `.getTimezoneOffset()` from the date object as part of the stored shape, so even if it gets rehyrdated in a different zone you could still have some idea of what the local time was when generating it. But, that feels perhaps clunky to me.

PR #140, opened 2018-12-06
CRDT increment operation

ept:
A counter is the most frequently cited and most trivial example of a CRDT, because addition of integers is naturally commutative. So far, Automerge hasn't supported counters, since we haven't actually needed them, but it does seem like a missing feature that ought to be there.

So, in theory, a counter is trivial. In practice, some interesting questions arise, such as: How do we handle the case where one user assigns a conflicting value (that may not even be a number) concurrently to another user incrementing a number? And, how do we handle undo of an increment (do we produce a matching decrement operation, or do we reset to the old value)?

Moreover, there an important API design question arises: how does the application indicate that it wants to increment a value, as opposed to overwriting the old value with a new value? One option would be to introduce a very explicit API, for example:

```js
after = Automerge.change(before, doc => {
  Automerge.increment(doc, 'counter', 1) // increases the value of doc.counter by 1
})
```

…but that feels quite ugly to me. I would like to be able to write the following code, and have it perform a CRDT increment operation, because that's clearly the intent:

```js
after = Automerge.change(before, doc => {
  doc.counter++
})
```

However, the JavaScript [Proxy](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy) feature that we use to capture modifications in the change callback does not allow us to override the `++` or `+=` operators, so from the point of view of Automerge, the code above is indistinguishable from:

```js
after = Automerge.change(before, doc => {
  doc.counter = doc.counter + 1
})
```

The proxy *can*, however, detect that the counter was first read, and then written. Thus, we can distinguish between the code above and the following code:

```js
oldValue = doc.counter
after = Automerge.change(before, doc => {
  doc.counter = oldValue + 1
})
```

In this last example we can detect that the counter was written but not read within the change callback. We can't detect any reads outside of the change callback, so the last example is indistinguishable from just assigning any numeric constant to the `counter` property.

So, the implementation in this PR takes the following approach:

* If a field has a numeric value, and it has no conflicts, and it is read within the change callback, and it is assigned another number value within that change callback, then the write is treated as an increment (where the increment magnitude is the difference between the new and the old value).
* In all other cases, the assignment is treated as an overwrite of the previous value.

This has the advantage that code like the following produces a commutative increment operation:

```js
after = Automerge.change(before, doc => {
  doc.counter += 5
})
```

while the following code produces a regular assignment operation:

```js
after = Automerge.change(before, doc => {
  doc.positionX = 102
  doc.positionY = 321
})
```

This seems nice to me, but it runs the risk of being a bit too "magic". While I think the above heuristic will work in the common case, there may be situations where it guesses wrongly. For example, the following code (which snaps the x/y position of a graphical object to the nearest multiple of 100 pixels) should use assignment, but it incorrectly guesses that you want to increment the x and y coordinates, since each coordinate is first read and then written:

```js
after = Automerge.change(before, doc => {
  doc.positionX = Math.round(doc.positionX / 100) * 100
  doc.positionY = Math.round(doc.positionY / 100) * 100
})
```

Guessing the wrong operation type will have undesirable merge semantics (the when merged, objects will most likely no longer be snapped to the 100 pixel grid). However, if your code produces an increment when you wanted an assignment, or vice versa, the mistake is not very visible: assignment and increment behave identically in the absence of concurrency, and only differ when there are concurrent changes.

So, how do we design the API to minimise the risk that application developers end up with the wrong semantics? Even introducing an explicit API, like in the first code example above, does not guarantee that people will actually use it; I'm sure people will be very tempted to just write `+= 1`, no matter how much we document that doing so is unsafe. So my inclination is to try to make the `+= 1` case safe by guessing that it should be an increment. For cases like the snapping-to-grid example, which I think will be rarer, the workaround would be to perform the calculation outside of the change callback, and to perform an unconditional assignment within the change callback:

```js
const positionX = Math.round(before.positionX / 100) * 100
const positionY = Math.round(before.positionY / 100) * 100
after = Automerge.change(before, doc => {
  doc.positionX = positionX
  doc.positionY = positionY
})
```

Thoughts?

### DIFF
--- a/backend/op_set.js
+++ b/backend/op_set.js
@@ -176,7 +176,7 @@ function updateMapKey(opSet, objectId, key) {
   return [opSet, [edit]]
 }
 
-// Processes a 'set', 'del', or 'link' operation
+// Processes a 'set', 'del', 'link', or 'inc' operation
 function applyAssign(opSet, op, topLevel) {
   const objectId = op.get('obj')
   const objType = opSet.getIn(['byObject', objectId, '_init', 'action'])
@@ -191,11 +191,24 @@ function applyAssign(opSet, op, topLevel) {
     opSet = opSet.update('undoLocal', undoLocal => undoLocal.concat(undoOps))
   }
 
-  const priorOpsConcurrent = opSet
-    .getIn(['byObject', objectId, op.get('key')], List())
-    .groupBy(other => !!isConcurrent(opSet, other, op))
-  let overwritten = priorOpsConcurrent.get(false, List())
-  let remaining   = priorOpsConcurrent.get(true,  List())
+  const ops = opSet.getIn(['byObject', objectId, op.get('key')], List())
+  let overwritten, remaining
+
+  if (op.get('action') === 'inc') {
+    overwritten = List()
+    remaining = ops.map(other => {
+      if (other.get('action') === 'set' && typeof other.get('value') === 'number' &&
+          !isConcurrent(opSet, other, op)) {
+        return other.set('value', other.get('value') + op.get('value'))
+      } else {
+        return other
+      }
+    })
+  } else {
+    const priorOpsConcurrent = ops.groupBy(other => !!isConcurrent(opSet, other, op))
+    overwritten = priorOpsConcurrent.get(false, List())
+    remaining   = priorOpsConcurrent.get(true,  List())
+  }
 
   // If any links were overwritten, remove them from the index of inbound links
   for (let op of overwritten.filter(op => op.get('action') === 'link')) {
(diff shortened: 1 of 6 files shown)


j-f1:
Is there a way to temporarily modify `Number`’s prototype so people could do `doc.counter.set(42)` to disable the increment behavior?

ept:
After some discussion with others we concluded that this API (that tries to guess whether to assign or increment a value) is too "magic", and that a more explicit API would be preferable. I have implemented that second attempt on #147, and will close this one.

Issue #94, opened 2018-05-28
proposal: allow underscores


rikkertkoppes:
Currently underscores are not allowed in documents. This is unfortunate as it makes automerge less universal.

I understand this is to prevent accidental overwrites of tracking props, such as _actorId, _objectId, _conflicts and _state. When adding automerge to an existing application which heavily used underscores as part of another library, I ran into a conflict.

I think the best way to fix this is to use es6 symbols, or when symbols are not available, a configurable, namespaced property name like `@@automerge/state`

I'd be happy to propose a pull request, but I think it's best to discuss it a bit first

- are there any objections to this proposal?
- are there any other solutions that may come to mind?
- is the issue encountered by others? What were their solutions?

ept:
Hi @rikkertkoppes, thanks for your proposal. I agree that using ES6 symbols would be a nicer solution than reserving all underscore-prefixed names, and I would happily accept a pull request that does this.

Can the fallback from symbol to a namespaced property be done with reasonably clean code? Also, with this approach, what would the API for getting an object ID or conflicts look like? Would it be something like `Automerge.getConflicts(doc.foo)` instead of `doc.foo._conflicts`, or something else?

rikkertkoppes:
Working on it here: https://github.com/rikkertkoppes/automerge/tree/allow_underscore

Specifically for the fallback, I extracted a map of props here: https://github.com/rikkertkoppes/automerge/blob/allow_underscore/src/prop_names.js

Basically I replace `obj._objectId` with `obj[propNames._OBJECT_ID]` everywhere.

Indeed, for doc.foo._conflicts, we would need an api to replace it. I did not realize this, but this would be a breaking change. The thing is, I do not have much experience with automerge, so not really in a position to judge the impact.

What parts would break if all underscore-properties are removed? I'd like to get a list of that.

Maybe the approach would be to deprecate using the underscore props directly and expose all of them through an api like `Automerge.getConflicts(doc.foo)` This could be a first minor release

The next step (major release as it is a breaking change) is to replace all underscore props. Either with symbols, or with namespaced prop names (such as `@@automerge/objectId`)

I am still trying to find out which underscore props there actually are and what things break if I remove them. I do seem to have some problems with `_init` and `_inbound` when using them as symbols in conjunction with immutability. Still need to dive into that.

EthanRBrown:
[...]

PR #150, opened 2019-02-04
Allow property names starting with underscore

ept:
This PR implements the long-standing request #94: Automerge map objects may now contain user-defined properties whose name begins with an underscore. Previously these properties had been reserved for Automerge-internal purposes, but now applications may use them as they see fit.

In order to enable this change, all APIs that were provided through underscore-prefixed properties or methods on document objects have now been replaced with functions in the Automerge namespace. In particular:

* `doc.object._objectId` is now `Automerge.getObjectId(doc.object)`
* `doc.object._conflicts.property` is now `Automerge.getConflicts(doc.object, 'property')`
* `doc._get(objectId)` is now `Automerge.getObjectById(doc, objectId)`

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,11 +104,7 @@ const Automerge = require('automerge')
 // Further down we'll simulate a second device.
 let doc1 = Automerge.init()
 
-// That initial state is just an empty object: {}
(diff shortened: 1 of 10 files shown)


migueloller:
Thanks for this @ept!

Issue #142, opened 2018-12-15
Migrate to TypeScript


unao:
I think this project is super cool - thank you all for the hard work.

I wanted to suggest to consider migrating the code base to TypeScript, as it brings a lot of benefits and very little downsides. I did already initial work (~50-60% of all the work that is needed): https://github.com/unao/automerge/tree/ts and wanted to ask if it made sense to continue this path.

If there is some interest from the maintainers, I would happily discuss more details - btw. I totally understand if you prefer to stick to raw JS. In such a case I could probably help with: https://github.com/automerge/automerge/issues/105.

pvh:
Hi @telamon -- I'm interested to hear how you were using hypermerge. 

@unao, I think Automerge is likely to stay in JS but to acquire TS types. Really good and ergonomic types for Automerge are somewhat tricky because of the dynamic nature of the underlying documents (another node could send you unexpected data and violate your local types, for example), but if you look in Hypermerge at the fork branch I believe there's an automerge.d.ts file in there.

telamon:
@pvh I can't answer your question because i don't know what hypermerge does, it's source-code does not make sense to read without IDE assistance and/or the deluded notion of types in JS.

But it if you're on the other-hand are interested on what I'm doing in general, I suggest you drop me a pm to avoid further derailing this issue.
I'm still curious about how hypermerge is designed, just not curious enough to endure TS :)

schrepfler:
“deluded notion of types in JS” sounds a bit biased. IMO types are a good thing, especially in library design.

telamon:
@schrepfler of course it is biased, I hope you agree that having the opportunity to form your own opinion and bias is healthy and good for diversity?

I'm merely informing that converting a common library to sub-syntax-X is limiting it's overall compatibility and usefulness. And most importantly it hampers freedom of choice.

I couldn't care less what you choose to use in your own user-end project; you are entirely free to follow your own whims and I an fully defending your right to do so.

As for me, I have worked with many typed languages and dynamic languages for the last couple decades thus being coloured by my own experiences. So please excuse my liberal honesty in saying:
"Trying to solve complexity by adding another layer of complexity is just plain dumb"

Even if i have a different standing point - my intention is **not** to convert you, I'm merely asking you in a polite way to not pull your pants down in public spaces, and I sincerely promise to extend the same courtesy to you. :slightly_smiling_face:

pvh:
[...]

Issue #127, opened 2018-10-12
Configuring initial object on init()


samjohnduke:
Hi, 

Just getting started with learning about automerge and I am creating an exploratory application to have a go. 

I was looking at the documentation and it seem fairly straight forward but i was wondering if it makes sense to have a non empty root document when calling init. I would like to be able to say that my document conforms to a particular object (say {nodes: []}). 

Is this something that you think might be a good idea / even possible? 

Sam

pvh:
That sounds like a nice bit of syntactic sugar but I don't think needs any kind of special semantics. Would you agree?

HerbCaudill:
I'd also like to be able to initialize an automerge document with something other than `{}`. Currently you have to do this: 

```js
let state = Automerge.init()
state = Automerge.change(state, d => d.todos=[])
state = Automerge.change(state, d=> d.filter='ALL')
```

This would be a lot more readable:

```js
const initialState = {
  todos: [],
  filter: 'ALL',
}
let state = Automerge.init({initialState})
```

or even

```js
let state = Automerge.from({
  todos: [],
  filter: 'ALL',
})
```

I started to try to implement this, but didn't get very far - I think I need to understand a bit more of the internals. @pvh or @ept - any pointers you could provide to get me going in the right direction would be appreciated.

ept:
You can do this to set the initial state:

```js
let state = Automerge.change(Automerge.init(), doc => Object.assign(doc, initialState))
```

We could certainly consider adding a function `Automerge.from` that wraps up this pattern, e.g.:

```js
function from(initState) {
  return Automerge.change(Automerge.init(), doc => Object.assign(doc, initState))
}
```

…though I am not totally convinced that this one-liner is really worth adding.

There is also a deeper problem with initialisation that is discussed in [issue #4](https://github.com/automerge/automerge/issues/4#issuecomment-489569722): that is, initialisation does not happen only when a document is first created with `Automerge.init()`, but also at later points in its lifetime (e.g. when a new feature is added to the application, and that feature needs to initialize an empty object in the document to store its data). And that is the actual tricky issue. Initialisation on `Automerge.init()` is the easy bit.

HerbCaudill:
[...]

PR #93, opened 2018-05-24
Support replacing UUID implementation

kpruden:
Provide a way for replacing the UUID implementation at runtime.
This is useful for writing repeatable tests, and for contexts
where an alternative source of unique IDs is prefferred.

PR #155, opened 2019-04-09
TypeScript typings support

### LINKED ISSUE #105: Minor Request: Typescript typings support
As per the title. I was a little taken aback seeing that this such a good package has no typings yet. Though,  this is just a very minor request that doesn't really cause any issue for now.

HerbCaudill:
I've created a type definition file for Automerge. This fixes #105.

I understand from #142 that there's no appetite for converting the whole project to TypeScript. In order to ensure the correctness of the type definitions here, I've changed the *test files* to TypeScript. The source files are unchanged, with one exception described below [1]. 

If the team prefers *not* to change the test suite to TypeScript, an alternative to this PR would be to *only* add the `automerge.d.ts` file as-is. This would make it easier for the library to drift away from the type definitions, though. 

I haven't yet tested these typings in a dependent package; until I do that I'm leaving this as a draft PR. There are also a few TODOs to take care of. 

-----

[1] The only change to the source files was in `uuid.js`, where I removed the factory stuff and just re-exported the generator from the library. I spun my wheels for a while trying to get TypeScript happy with the factory pattern, and wasn't able to figure it out. Since we're not actually using the capability to swap out the UUID generator, I took the easy way out and just got rid of it.


ept:
Hi @HerbCaudill, thank you very much for this contribution!

Regarding the type definitions, could @pvh and/or @jeffpeterson please take a look and see how the definitions fit with Hypermerge and Farm?

Regarding the tests, it's hard to see how much of a difference the change from JS to TypeScript has actually made, because the files have been reformatted and all the whitespace has changed. Would it be possible to change JS to TS without reformatting? I also find the reformatted style tends to make lines very short, so functions use a lot of vertical space, making it more difficult to work on the code on smaller screens. I know Prettier is standard for formatting in a lot of TS projects, but I personally just don't like the Prettier style very much.

An alternative option might be to keep most of the tests in JS, and add some additional TS tests that are specifically designed to exercise the typed interface.

aslakhellesoy:
For the purpose of review, perhaps rename the .ts files back to .js temporarily so the diff is shown?

aslakhellesoy:
I've created https://github.com/automerge/automerge/pull/156 to make it easier to review this

HerbCaudill:
@alsakhellesoy I kept the renames and the type changes in separate commits, so you can diff the tests in commit 1ab136f .

@ept sorry about the reformatting - I hadn't noticed that I had Prettier running until I'd already made most of the changes. I'll see what I can do to make it easier to see the changes.

@types/automerge/index.d.ts
```diff
@@ -0,0 +1,253 @@
+declare module 'automerge' {
+  function applyChanges<T>(doc: T, changes: Change[]): T
+  function canRedo<T>(doc: T): boolean
+  function canUndo<T>(doc: T): boolean
+  function change<T>(doc: T, message: string, callback: ChangeFn<T>): T
```
bkniffler: This should be `message: string | undefined`.

```
let message = ifCondition ? 'Yay' : undefined;
const newDoc = Automerge.change(doc, message, (change: any) => {
              ...
});
```

HerbCaudill: ✅ fixed in 30cf6e

HerbCaudill:
@bkniffler I removed the UUID factory since it's not public and wasn't being used anywhere internally - from my original post:

> [1] The only change to the source files was in uuid.js, where I removed the factory stuff and just re-exported the generator from the library. I spun my wheels for a while trying to get TypeScript happy with the factory pattern, and wasn't able to figure it out. Since we're not actually using the capability to swap out the UUID generator, I took the easy way out and just got rid of it.

Let me know if you think that's a bad idea - I can give it another shot.

HerbCaudill:
@ept - I've amended the PR to remove the formatting changes, hopefully this makes it easier to review.

bkniffler:
@HerbCaudill Oh, I didn't notice that the generator is not being used. And there I was, proudly using https://github.com/ai/nanoid all the time, or at least I thought :) 

Would love to be able to swap. I'm using nanoid for most of my ID cases due to it using only 21 characters vs 36 in UUID/V4, faster and smaller in library size; and its got comparable uniqueness.

ept:
The ability to swap out the UUID generator was introduced in #93. It's not used in Automerge itself, but is intended for use in projects that build upon Automerge and want to control UUID generation (in particular for testing purposes).

HerbCaudill:
Ok. I'll restore that and figure out how to type it.

HerbCaudill:
@ept, @bkniffler - I've restored the UUID factory and added it to the type definitions. 

I've wrapped up the remaining TODOs on this PR, so it should be ready for review! 🎉

@types/automerge/index.d.ts
```diff
@@ -0,0 +1,275 @@
+declare module 'automerge' {
+  function applyChanges<T>(doc: T, changes: Change<T>[]): T
+  function canRedo<T>(doc: T): boolean
+  function canUndo<T>(doc: T): boolean
+  function change<T>(doc: T, message: string, callback: ChangeFn<T>): T
+  function change<T>(doc: T, callback: ChangeFn<T>): T
+  function diff<T>(oldDoc: T, newDoc: T): Diff
```
aslakhellesoy: @HerbCaudill I believe the return type here should be `Diff[]`

@types/automerge/index.d.ts
```diff
@@ -0,0 +1,275 @@
...
+    registerHandler(handler: Handler<T>): void
+    unregisterHandler(handler: Handler<T>): void
+  }
+
+  class WatchableDoc<T> {
+    constructor(doc: T)
+    applyChanges(changes: Change<T>[]): T
+    get(): T
+    set(doc: T): void
+    registerHandler(handler: Handler<T>): void
```
aslakhellesoy: The `handler` argument here should be a function with signature `(doc: T) => void` (there is no `UUID` argument).

I suggest renaming `Handler` to `DocSetHandler` and introducing a new type `WatchableDocHandler`.

/cc @HerbCaudill

ept:
Thanks for your patience, I am looking into this now. I think my preferred approach is to keep the existing tests in plain JS as well, and to add new TS tests that are specifically designed to catch any divergence between the types and the implementation.

ept:
Trying to run `yarn test` locally I now get the following error:

```
/Users/martin/dev/cl/automerge/node_modules/transit-js/transit.js:2582
      a(this._entries[b + 1], this._entries[b]);
      ^
Error: Cannot create handler for JavaScript undefined
    at /Users/martin/dev/cl/automerge/node_modules/transit-js/transit.js:3492:15
    ....
```

Anyone have any ideas?

ept:
With @HerbCaudill's generous help I was able to fix the above exception that was preventing me from running the tests locally. (The problem was that in my Automerge repo directory I had a local file called `immutable/index.js`, which was not added to git, and the transit-js package was loading this file instead of the NPM "immutable" package when requiring `immutable`. The file had been there for some time, and I'm not sure why this problem happened to show up now; perhaps importing TypeScript changed the file path resolution logic for requiring packages.)

I have now reverted the existing tests to being plain JavaScript, and I am adding a separate set of tests that specifically test the type definitions. In this context I have another question for people more familiar with TypeScript:

I would like to test that the observed types at runtime match those that are declared. For example, `Automerge.save` and `Automerge.load` could be declared to have the following types:

```typescript
function save<T>(doc: T): string
function load<T>(doc: string, actorId?: string): T
```

so the following code type-checks correctly as expected:

```typescript
Automerge.load(Automerge.save(doc))
```

However, if I change the type declaration of `save` to return `number` instead of `string`, and change the first parameter of `load` to take `number` instead of `string`, then `Automerge.load(Automerge.save(doc))` still type-checks, even though the type declaration is wrong (`save` does indeed return a string, not a number).

Is there some way I can get TypeScript to throw a runtime error if the type observed at runtime does not match the declared type, in order to catch such mistakes in the type declaration?

bkniffler:
@ept Not sure if I got you completely right, but Typescript does not runtime type checking. Types are completely stripped out during compilation. So it makes sense that `Automerge.load(Automerge.save(doc))` works if you change the return type of save to number and first argument of load too. You won't be able to do something like:
```
let data = Automerge.save(doc);
saveTextToDatabase(data); // will throw a type warning during compilation if data is number 
data = loadTextFromDatabase()
Automerge.load(data)
```
Assuming that `saveTextToDatabase` and `loadTextFromDatabase` expect and return string. While this example might seem stupid, it is way more probable that in a real project you will somehow store the result into objects and variables that are expected to be strings and thus getting a compilation error.

So maybe changing the test to something like:
```
const data: string = Automerge.save(doc);
Automerge.load(data)
```
... might help make sure we are really handling strings between saving and loading.

HerbCaudill:
@ept - @bkniffler is right - there's no such thing as runtime Typescript, but your tests _should_ fail if there are type violations. The snippets you posted should be doing that, so I'm not sure what the problem is. If you want to push your work in progress, I'd be happy to take a look.

ept:
Ok, no worries. Thanks for replying.

I have put some additional commits on the [types branch of this repo](https://github.com/automerge/automerge/compare/types) (I don't know if I can add them to this PR?), including a set of typescript-specific tests and a few corrections to the type definitions. I plan to add some more tests to make sure the tests cover most of the API surface area of the public API, but the internals (such as `skip_list.js`) don't need TypeScript tests, since they are not exposed to applications using Automerge.

@types/automerge/index.d.ts
```diff
@@ -0,0 +1,275 @@
...
+  function equals<T>(val1: T, val2: T): boolean
+  function getActorId<T>(doc: T): UUID
+  function getChanges<T>(oldDoc: T, newDoc: T): Change<T>[]
+  function getConflicts<T>(doc: T, key: Key): any
+  function getHistory<T>(doc: T): State<T>[]
+  function getMissingDeps<T>(doc: T): Clock
+  function getObjectById<T>(doc: T, objectId: UUID): T
+  function getObjectId<T>(doc: T): string
+  function init<T>(actorId?: string): T
+  function load<T>(doc: T, actorId?: string): T
```
aslakhellesoy: Should be:
@HerbCaudill this should be:

```typescript
function load<T>(s: string, actorId?: string): T
```

@types/automerge/index.d.ts
```diff
@@ -0,0 +1,275 @@
...
+  function merge<T>(localDoc: T, remoteDoc: T): T
+  function redo<T>(doc: T, message?: string): T
+  function save<T>(doc: T): T
```
aslakhellesoy: @HerbCaudill this should be:

```typescript
function save<T>(doc: T): string
```

HerbCaudill:
@ept Looks to me like everything is working properly - if I change the return value of `save` to `number`, the tests fail, which is what we want. 

image

HerbCaudill:
btw you should be able to commit changes directly to this PR if you want:

airhorns:
Anything the community could help with to move this forward?

pvh:
Reports from users that the types look good and meet your needs are helpful for sure. It's one of those things that is painful to change once they've been adopted in the field.

eldh:
I'm trying them out now and it's working well so far. Haven't used it for that long though. Will report back in a while.

ept:
I think we're very close. Some experience reports, as @pvh suggested, would be helpful. We also need some tests for the TypeScript type definitions for `Text`, `Table` and `Counter` (see 384865ee9be163595c94536f4c1bff0e9d57868c), which could be added by a volunteer; then I think we're done.

I am currently rushing to finish a paper for a deadline in a few days, but I should be able to return to this next week.

HerbCaudill:
Thanks, Martin. I'll add those tests.

airhorns:
I've been using the Automerge typings from this branch in a project with a non-trivial document type and haven't run into any issues, functional or ergonomic. The only one that's maybe a little weird is when initializing a document. 

```typescript
interface DocType { 
  key: boolean;
  otherKey: string;
}

const doc = Automerge.init<DocType>()
``` 

You have to be explicit about the type when creating the document in order to get the right return type from `init`. I'm not really sure there's a way around that and it's certainly not a big issue, just potentially confusing for new users of TypeScript + Automerge.

HerbCaudill:
@ept I've added tests for `Text`, `Table`, and `Counter` - let me know if there's anything else we need to do to get this merged! 

I currently have copies of `automerge/index.d.ts` all over the place, looking forward to getting those directly from the package. 😁

HerbCaudill:
@airhorns - You can also do this: 
```typescript
const doc: DocType = Automerge.init()
```

I've added a test for this usage, along with some notes, here:
https://github.com/automerge/automerge/blob/8c9b503781b7cb4aaeb89f0378a8bbc05edd6624/test/typescript_test.ts#L27-L50

airhorns:
@HerbCaudill that makes sense, but the reason I was doing it the other way is because I (falsely) assumed that the return type from `Automerge.init` would be something like `DeepReadonly<DocType>` or some other such fancy type that I probably shouldn't import and couple to. I think it might be good to have the types reflect the semantics of Automerge and prevent you from modifying the document outside the change block, no?

HerbCaudill:
@airhorns - 

> I think it might be good to have the types reflect the semantics of Automerge and prevent you from modifying the document outside the change block, no?

I guess I wouldn't want the typings to be any more or less opinionated about that kind of thing than the library itself. Automerge could return immutable objects, but it doesn't. 

It's not too hard to imagine a scenario where you got something back from Automerge and then wanted to modify it further, for whatever reason maybe whatever you're doing no longer needs to be captured by Automerge.

airhorns:
Huh, I must be missing something, cause the readme seems to indicate that the doc is indeed immutable outside of the change block via `Object.freeze`. From the readme: 

```js
// Let's say doc1 is the application state on device 1.
// Further down we'll simulate a second device.
let doc1 = Automerge.init()

// That initial state is just an empty object: {}.
// The doc1 object is immutable -- you cannot change it directly (if you try,
// you'll either get an exception or your change will be silently ignored,
// depending on your JavaScript engine). To change it, you need to call
// Automerge.change() with a callback in which you can mutate the state. You
// can also include a human-readable description of the change, like a commit
// message, which is stored in the change history (see below).

doc1 = Automerge.change(doc1, 'Initialize card list', doc => {
  doc.cards = []
})
```

And in my use of Automerge I've been bitten by the fact that those objects are indeed frozen and do indeed have the heisenbugs of sometimes throwing TypeErrors or not when you accidentally mutate them outside a change block.

HerbCaudill:
@airhorns you're right, my mistake!

So yeah, the return type probably should be `DeepReadOnly<T>`. Let me play with that a bit.

ept:
@HerbCaudill The new tests look great, thank you. I have also added types for `Automerge.from` (#127). I am happy to merge and release this now if you think it's ready.

Regarding immutability, yes, Automerge does `Object.freeze` on document objects to prevent inadvertent mutations. It might be nice to reflect this immutability in the type, but I don't know how much complexity that adds. You can let me know whether you think using a `DeepReadOnly<T>` type would be worthwhile.

HerbCaudill:
Thanks, @ept. I've gone a little ways down the road of making the return object types read-only and it looks like a non-trivial undertaking. (Details follow). It would be nice to have compile-time protection against the kind of heisenbug @airhorns describes, but we can implement this later if we figure out how. For now I'd say let's go with what we have. 

Cheers!
Herb

### Details FWIW

Documenting this in case someone understands the issues better than I do & can suggest an easier route.

Typescript has a `readonly` modifier for individual properties, but deep-freezing arbitrary objects happens in userland. The simplest implementation I've found looks like this:
```typescript
type DeepFreeze<T> = 
  T extends Function ? T :
  T extends ReadonlyArray<infer R> ? DeepFreezeArray<R> : 
  T extends Map<infer K, infer V> ? DeepFreezeMap<K, V> : 
  T extends object ? DeepFreezeObject<T> : 
  T

interface DeepFreezeArray<T> extends ReadonlyArray<DeepFreeze<T>> {}
interface DeepFreezeMap<K, V> extends ReadonlyMap<DeepFreeze<K>, DeepFreeze<V>> {}
type DeepFreezeObject<T> = { readonly [P in keyof T]: DeepFreeze<T[P]> }
```

So for example we can indicate that the result of `init` is frozen: 
```typescript
function init<T>(options?: any): DeepFreeze<T>
```

The problem comes in the `change` callback, when the object needs to be unfrozen but isn't:
```typescript
let s1 =  Automerge.init<BirdList>() // s1 is DeepFreeze<BirdList>
s1 = Automerge.change(s1, doc => (doc.birds = ['goldfinch'])) // Cannot assign to 'birds' because it is a read-only property.ts(2540)
``` 

I thought it might work to modify the signature of `change` to indicate that it takes a frozen object and returns a frozen object, but that the callback parameter isn't frozen:
```typescript
function change<T>(doc: DeepFreeze<T>, callback: ChangeFn<T>): DeepFreeze<T>
type ChangeFn<T> = (doc: T) => void
```

This is conceptually the right route, evidenced by the fact that this make some of the errors go away:
```typescript
function change<T>(doc: DeepFreezeObject<T>, callback: ChangeFn<T>): DeepFreeze<T>
```
But I think backing out `DeepFreeze` just goes beyond Typescript's inference capabilities. 

An alternative would be to make our own `DeepThaw` to back it out, something like this:
```typescript
function change<T>(doc: T, callback: ChangeFn<DeepThaw<T>>): T
```

¯\\\_(ツ)\_/¯

airhorns:
> I thought it might work to modify the signature of change to indicate that it takes a frozen object and returns a frozen object, but that the callback parameter isn't frozen:

I think you're right and this is what would need to happen.

> But I think backing out DeepFreeze just goes beyond Typescript's inference capabilities.

I think that's on the money. The issue is that the `infer` keyword that the `DeepFreeze` implementation uses doesn't work for forward and backwards inference, I think.. If you use something like `Readonly` which doesn't do any internal inference, the problem mostly goes away, I tried. I think that if we want to reflect the readonly semantics with the types `DeepThaw` would be necessary, or, someone has to work some serious incantations to get TypeScript to understand how to infer the inner frozen type.

HerbCaudill:
@airhorns - strong work! I took a look at your definitions and it occurred to me that the root Automerge document is always an object, so the frozen version should be `Doc<T> = DeepFreezeObject<T>` and not `DeepFreeze<T>`. With that, TypeScript is able to magically infer everything it needs to. 

I only had to tweak a couple of tests that were overspecifying the output of `init`, and upgrade Typescript to get the `Omit` helper function. 

This is good to go - tests pass, works beautifully in my projects. 

@ept plz merge before I spend any more time on this!! 😁

aslakhellesoy:
I think I have found a few incorrect signatures. It seems to me the `Backend` namespace should declare the `state` parameter as `State<T>` rather than `T`. I'm working on a PR for this. I'd recommend holding off on merging this until I've got the PR ready.

I discovered these errors while porting the `Connection` class to TypeScript. My motivation for doing this is to refactor it to a `WatchableDocConnection` as an alternative to `Connection`.

ept:
@aslakhellesoy I would even go a step further, and say that the Backend state should be simply an opaque `BackendState` type, without any type parameter. As far as I remember, the backend API does not expose any functions or properties that depend on the shape of the document; it only consumes or returns `BackendState`, `Change`, `Patch` and `Clock` objects. I have made this change in a20bec87f184d5788f46080bb59502921445bdd4.

Previously, `Change<T>` was parameterised by `T` because it contained a field `before` of type `T`. I think this field should not be considered part of the public API (it is used for internal bookkeeping within the frontend, but the `Change` objects produced and consumed by the backend do not have this property). In 55ffba67dfdbf1f8632e5a0197def744fe6d8eab I have removed the `before` property, and removed the type parameter from `Change`.

@aslakhellesoy Does this fit what you had in mind?

Final call for any changes to this PR before I merge and release it? (We can still make changes in subsequent versions, but any changes that would cause compatibility problems to TypeScript users might best be made now.)

aslakhellesoy:
Hi @ept - that sounds good to me! I won't have a chance to verify that it works for me until next week, but just go ahead and merge/release this. If it's still a bit broken in some of the obscure areas of the API I don't think it will be too disruptive to push a fix in a subsequent release.

ept:
I have now merged this PR, and released Automerge 0.11.0 to npm. Could you please check in your TypeScript projects that the type definitions are getting picked up as expected?

Huge thanks to @HerbCaudill, @airhorns, @aslakhellesoy, @bkniffler and everyone else who contributed to this. It was a big piece of work and I am very pleased how it has come together. This has been an example of open source at its best. :)

PR #163, opened 2019-05-16
Nicer JSON serialization of Counter, Table, and Text

ept:
Before:

```js
JSON.stringify(Automerge.change(Automerge.init(), doc => doc.counter = new Automerge.Counter()))
// {"counter": {"value": 0}}

JSON.stringify(Automerge.change(Automerge.init(), doc => {
  doc.blog = new Automerge.Table(["author","title"])
  doc.blog.add({author: 'martin', title: 'hello'})
}))
// {"blog": {
//   "entries": {
//     "columns": ["author","title"],
//     "57a47ad9-b273-4657-b871-2a8341bd4218": {"author": "martin", "title": "hello"}
//   },
//   "columns":["author","title"]}
// }

JSON.stringify(Automerge.change(Automerge.init(), doc => {
  doc.text = new Automerge.Text()
  doc.text.insertAt(0, 'a', 'b', 'c')
}))
// {"text": {"elems": [
//   {"elemId": "85ec7c04-979c-4db0-9133-48d61cec0b67:1", "value": "a"},
//   {"elemId": "85ec7c04-979c-4db0-9133-48d61cec0b67:2", "value": "b"},
//   {"elemId": "85ec7c04-979c-4db0-9133-48d61cec0b67:3", "value": "c"}
// ]}}
```

After:

```js
JSON.stringify(Automerge.change(Automerge.init(), doc => doc.counter = new Automerge.Counter()))
// {"counter": 0}

JSON.stringify(Automerge.change(Automerge.init(), doc => {
  doc.blog = new Automerge.Table(["author","title"])
  doc.blog.add({author: 'martin', title: 'hello'})
}))
// {"blog": {
//   "columns": ["author", "title"],
//   "4485409b-7242-4019-97f0-27ebccf26996": {"author": "martin", "title": "hello"}
// }}

JSON.stringify(Automerge.change(Automerge.init(), doc => {
  doc.text = new Automerge.Text()
  doc.text.insertAt(0, 'a', 'b', 'c')
}))
// {"text": "abc"}
```

### DIFF
--- a/frontend/counter.js
+++ b/frontend/counter.js
@@ -29,6 +29,14 @@ class Counter {
   toString() {
     return this.valueOf().toString()
   }
+
+  /**
+   * Returns the counter value, so that a JSON serialization of an Automerge
+   * document represents the counter simply as an integer.
+   */
+  toJSON() {
+    return this.value
+  }
 }
 
 /**
--- a/frontend/table.js
+++ b/frontend/table.js
@@ -200,7 +200,9 @@ class Table {
    * when serializing an Automerge document to JSON.
    */
   toJSON() {
-    return Object.assign({}, {columns: this.columns}, this.entries)
+    const rows = {}
+    for (let id of this.ids) rows[id] = this.byId(id)
+    return {columns: this.columns, rows}
   }
 }
 
--- a/frontend/table.js
+++ b/frontend/table.js
@@ -193,6 +193,15 @@ class Table {
     instance.entries = this.entries
     return instance
   }
+
+  /**
+   * Returns an object containing both the table entries (indexed by objectID)
+   * and the columns (under the key `columns`). This provides a nice format
+   * when serializing an Automerge document to JSON.
+   */
+  toJSON() {
+    return Object.assign({}, {columns: this.columns}, this.entries)
+  }
 }
 
 /**
--- a/frontend/text.js
+++ b/frontend/text.js
@@ -30,6 +30,14 @@ class Text {
       }
     }
   }
+
+  /**
+   * Returns the content of the Text object as a simple string, so that the
+   * JSON serialization of an Automerge document represents text nicely.
+   */
+  toJSON() {
+    return this.join('')
+  }
 }
 
 // Read-only methods that can delegate to the JavaScript built-in array
--- a/test/frontend_test.js
+++ b/test/frontend_test.js
@@ -208,6 +208,11 @@ describe('Frontend', () => {
         assert.strictEqual(`I saw ${doc1.birds} birds`, 'I saw 3 birds')
         assert.strictEqual(['I saw', doc1.birds, 'birds'].join(' '), 'I saw 3 birds')
       })
+
+      it('should allow counters to be serialized to JSON', () => {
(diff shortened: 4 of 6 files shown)


nthsense:
This is looking good!  The only thing that seemed unexpected to me was the structure of the Table json.  It looks like you're using a GUID for the row data.  It might make sense to have that be a `rows` array, where each row is an object.  This would make it easier to iterate over the rows, and they would all be at a predictable key.  However, I understand that `Table` is unordered data, and if you wanted to recreate the Table from the JSON, you'd need those IDs...  `_guid` could be an attribute on every row.

As this stands, if I want to do something like sort my rows by `"lastName"`, I would first have to transform the JSON which is a bit unwieldy.  I would conceptually do something like `Object.keys( myTable ).filter( key => key != "columns" ).map( key => myTable[key] ).sort( ... )`.

Also, I don't know if it's possible to have a table with column names that are not valid JSON keys - if so, that could cause problems in creating the row JSON.

Anyway - I don't understand the `Table` use-case nearly as much as you do, so I could be off-base.  I don't have an immediate need to use the `Table` JSON, I'm just coming at this from a conceptual level.

Thanks!

ept:
Thanks @nthsense for your comments. The idea of using the unique ID of each row as the key is an idea I [took from the Redux documentation](https://redux.js.org/recipes/structuring-reducers/normalizing-state-shape), which has recommendations on representing relational-like data in JavaScript.

I'm not too keen on making the rows an array, because the order would be meaningless, and I think the API should convey that fact. Also, I'm not keen on adding a `_guid` attribute to each row, since the application might create an attribute of the same name, which would clash (since #150 we no longer reserve attributes beginning with underscore for Automerge internally).

A proposal: we move the rows under a key `rows`, so that the table object has just two attributes: `rows` and `columns`. Under `rows` we keep the rows in the current form (mapping UUID to row object). To iterate over the rows, you can then use `Object.values(doc.table.rows)`.

pvh:
This makes sense to me! Nice bit of work, @ept.

nthsense:
@ept I see where you're coming from now.  Not being a Redux user, I hadn't encountered this pattern before, so I found it surprising.  In React apps, I always make sure the "key" on a React element is pointing at an attribute in the object that is unique within that set.  That way, React can recycle the element.  This is a pretty common pattern, at least from what I've seen.  (The Redux documentation you pointed to actually recommends a library called "Normalizr" to convert from the common format to the Redux-preferred format.)   Given the "normalized" data structure, I would still need to map over the keys and insert that guid into the object myself (unless I want to use Redux).  For small Tables, that won't be enough of a performance impact to worry about, and it does avoid the problem of name collisions by putting the control in the hands of the developer.

> the order would be meaningless

That's true - but not a bad thing.  This is analogous to "natural order" when querying a database.  Unless you specify a sort, the order is undetermined, and that's expected behavior.

All that said, I think your proposal is a good compromise.  If reshaping the data proves to be too taxing, it can be revisited in future.

ept:
Okay, I have changed `Automerge.Table.toJSON()` as proposed, and merged this PR. Thanks for your input!

PR #169, opened 2019-06-01
Add dist dir from build

canadaduane:
Automerge needs built artifacts for import in pushpin#upgrade:

```
  ERROR in ./src/renderer/components/text-content.js
  Module Error (from ./node_modules/eslint-loader/index.js):

    ✘  https://google.com/#q=import%2Fno-unresolved  Unable to resolve path to module 'automerge'
    src/renderer/components/text-content.js:6:23



  ✘ 1 problem (1 error, 0 warnings)
```


ept:
Hi @canadaduane, I'm not very keen on checking transpiled code into Git, because it messes up the commit history and makes `git grep` a lot less useful. Can you explain a bit more about why you think this is needed? Thanks.

canadaduane:
Sorry for the sparse explanation. I'm not sure what the current state of things in pushpin is (I've been away a week on vacation) but the "upgrade" branch used to require a pinned github-based version of automerge (rather than the npm package) due to depending on changes in automerge more recent than published. In order to get that to work, however, the `dist` directory had to exist. I think this is just a temporary solution, so you can probably ignore. @pvh would be better able to speak to the current state. Having a fork with `dist` checked in was a sufficient band-aid for me.

PR #171, opened 2019-06-07
Detect the current platform's behaviour for throwing on frozen object modification for the tests

airhorns:
Different platforms sadly treat modifying a frozen object differently, so this test has to test different behaviour depending on the platform. Before this change, it hardcoded which behaviour to expect when based on the presence of the window object, but new node versions changed their behaviour to match that of the browsers, breaking this check. Instead of a hardcoded check, let's just detect the current platform's behaviour outside the test, and then make sure that automerge matches that behaviour.

This fixes the tests to run cleanly on node v10.x,  v11.x, and v12.x. It also adds v11 and v12 to be tested on Travis to prevent this in the future!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,8 @@ node_js:
 - '8'
 - '9'
 - '10'
+- '11'
+- '12'
 script: 'yarn test && yarn build && TEST_DIST=1 yarn test && node -e "const Automerge = require(\"./dist/automerge\")"'
 jobs:
   include:
--- a/test/test.js
+++ b/test/test.js
@@ -42,10 +42,17 @@ describe('Automerge', () => {
         assert.deepEqual(s2, {first: 'one', second: 'two'})
       })
 
+      let platformIgnoresFrozenMutation = true;
+      try {
+        Object.freeze({}).foo = 1;
+      } catch (e) {
+        platformIgnoresFrozenMutation = false;
+      }
+
       it('should prevent mutations outside of a change block', () => {
         s2 = Automerge.change(s1, doc => doc.foo = 'bar')
-        if (typeof window === 'object') {
-          // Chrome and Firefox silently ignore modifications of a frozen object
+        if (platformIgnoresFrozenMutation) {
+          // Chrome, Firefox, and node >= 10 silently ignore modifications of a frozen object
           s2.foo = 'lemon'
           assert.strictEqual(s2.foo, 'bar')
           const deleted = delete s2['foo']


HerbCaudill:
@airhorns - My understanding is that this isn't a matter of which platform is being used, but whether we're in strict mode or not. A better solution is to just `use strict` everywhere.

See [this commit](https://github.com/automerge/automerge/pull/155/commits/22de4c89e31c45ed958fa09b564ad04c6307b67c) from #155 . 

> The original code here suggests that in a browser, modifying a frozen object will fail silently, whereas
in node it will throw an error. This test was failing for me in browser testing, so I dug into it and it turns out the difference isn't node vs browser but strict mode vs not. (In mocha.opts we had "use strict" so node tests were always running in strict mode. TypeScript was also compiling in strict mode.)
>
> Since we have control over whether we're in strict mode or not, it makes more sense to just turn it on everywhere and only test for that scenario.

airhorns:
Ah -- thanks for the extra context. Despite the mocha opts being set to turn on strict mode, the test in question still fails on node > 10 and doesn't on node <= 10 which makes me think that there might be another factor at play. I agree that using strict mode everywhere is ideal and your branch from #155 passes on the newer node versions as well. 

That said -- it might be good to still test automerge's behaviour if the invoking code is not in strict mode, because the library can't actually control that right? I am not deeply familiar with the strict mode semantics but I imagine that there will be "userland" code in change blocks or what have you living in files that aren't in strict mode. If that's true then I think the two branches of this test both have value but if not we can close this PR and you/me can land the change to the travis config after #155 gets merged.

ept:
I prefer a simpler approach to fixing these tests, as implemented in d00aa7d85c543ae539185d3035583313b932826f: don't care whether or not the mutations of frozen objects throw an exception, and just check that the mutations don't take effect.

However, turning on strict mode seems like a good idea in general. Do I have to add `'use strict';` to the top of every file? I notice that in the generated `dist/automerge.js`, Webpack already seems to add `'use strict';` to every module, so maybe the declaration in each file is not necessary?

airhorns:
Ok I will close this and open a new PR to add the other node versions to the test matrix. 

However, it was rather surprising to me that when I was modifying something accidentally outside a change block that the block didn't throw an exception and my changes were just lost. For people following Redux or other architectures along those lines there can be lots of indirection where it's hard to know if you are inside a change block or not. I think for this reason it'd be good for Automerge to try to throw as many exceptions as possible which is in line with what Martin said.  think the principle of least surprise says to me that modifying frozen objects should definitely throw an exception. [ES2015 modules are always strict](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode#Strict_mode_for_modules) which might be why Webpack is adding the strict mode declarations to the output and would indicate that you don't need to add it in the actual source.

ept:
Agree that it is desirable to throw an exception when you modify a frozen object. However, if I understand the scoping rules for strict mode correctly, you need to enable strict mode in your *application code* for this to happen; enabling strict mode in Automerge internals will not make any difference here.

airhorns:
I believe that's correct. Perhaps worth mentioning in the readme near the bit talking about why docs are frozen?

PR #175, opened 2019-06-26
add `Automerge.from`

HerbCaudill:
Returns a new document object with the given initial state. 
```js
const foo = Automerge.from({
  birds: {
    wrens: 3,
    magpies: 4
  }
})
```

Solves #127

### DIFF
--- a/frontend/index.js
+++ b/frontend/index.js
@@ -231,6 +231,15 @@ function init(options) {
 }
 
 /**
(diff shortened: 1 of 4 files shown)


ept:
Thanks! I found one bug (0e7ce7c0fdd82f5f088c30e1ba06bb3ffc681c9b has a failing test case) but it was easy to fix.

PR #178, opened 2019-06-29
DocSet tests

HerbCaudill:
Noticed that DocSet does not have code coverage. Added tests based on the WatchableDoc tests.


ept:
Thank you @HerbCaudill!

PR #179, opened 2019-07-03
Make use of Object.freeze optional to improve performance

ept:
As discovered by @izuchukwu in #177, our use of `Object.freeze` has a big negative performance impact on Automerge. In particular, in Chrome and Node (i.e. V8), copying a frozen array is [awfully slow](https://jsperf.com/cloning-frozen-array/1) (on the order of 100x slower than copying a non-frozen array of the same size), and copying a frozen object [isn't great either](https://jsperf.com/cloning-frozen-object/1). I have [filed a Chromium bug](https://bugs.chromium.org/p/chromium/issues/detail?id=980227) that will hopefully cause this issue to be fixed, but we don't know how long it will take.

In the meantime, this PR disables the use of `Object.freeze` by default. It doesn't remove all calls to that function, but it does remove those that are on the critical path performance-wise. If you want to keep the use of `Object.freeze`, e.g. in order to avoid inadvertently modifying an Automerge document outside of `Automerge.change()`, you can pass in an option when constructing the document object: `Automerge.init({freeze: true})` or `Automerge.load(string, {freeze: true})`.

With these changes, Automerge achieves the same performance as the "hot-automerge" fork in @izuchukwu's [benchmark](https://github.com/izuchukwu/experiments-automerge).

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,11 +55,7 @@ and merging**:
   point in time. Whenever you make a change, or merge in a change that came from the network, you
   get back a new state object reflecting that change. This fact makes Automerge compatible with the
   functional reactive programming style of [Redux](http://redux.js.org/) and
-  [Elm](http://elm-lang.org/), for example. Internally, Automerge is built upon Facebook's
-  [Immutable.js](http://facebook.github.io/immutable-js/), but the Automerge API uses regular
-  JavaScript objects (using
-  [`Object.freeze`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/freeze)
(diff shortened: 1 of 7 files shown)


lightsprint09:
This is a nice improvement. Speeds up loading docs about 100-150% using Apples JavaScriptCore.

What about `transit.fromJSON(string)` is this needed anymore? Does it have performance impact?

ept:
I have merged this and released it as part of Automerge 0.11.0.

@lightsprint09: We have plans to get rid of transit and replace it with a more efficient encoding, but that will have to happen separately — it's got nothing to do with object freezing.

Issue #180, opened 2019-07-12
Automerge.Text does not have documented insertAt / deleteAt methods


Gozala:
[According to readme](https://github.com/automerge/automerge/blob/a6c74e7c7e088736b9fb9c9f2004b04706a092ca/README.md#text-editing-support) `Text` type should have `.insertAt` / `.deleteAt` methods but as far as I can tell it [has neither](https://github.com/automerge/automerge/blob/a6c74e7c7e088736b9fb9c9f2004b04706a092ca/frontend/text.js)

It's also pretty unclear to me how one would make any changes to the `Text` blocks.

Gozala:
Oh it appears there is a proxy magic makes it so following works:

```js
Automerge.change(Automerge.init(), doc => {
  doc.text = new Automerge.Text()
  doc.text.insertAt(0, 'h', 'e', 'l', 'l', 'o')
  doc.text.deleteAt(0)
  doc.text.insertAt(0, 'H')
})
```

While following

```js
Automerge.change(Automerge.init(), doc => {
  const text = new Automerge.Text()
  text.insertAt(0, 'h', 'e', 'l', 'l', 'o')
  text.deleteAt(0)
  text.insertAt(0, 'H')

 doc.text = text
})
```
throws exception:

```
Uncaught TypeError: text.insertAt is not a function
    at Automerge.change.doc (<anonymous>:3:8)
    at Object.change (index.js:261)
    at Object.change (automerge.js:26)
    at <anonymous>:1:11
```

ept:
Yes, that's correct. At the moment, Automerge.Text objects are immutable by default; they gain their mutability only by being part of a document within a change callback. See also #166. 

However, I understand that this behaviour is confusing. I'll look into changing it. Allowing mutations of a Text object before it is assigned to a document will also enable #170.

Gozala:
I wrote a patch for #166 https://github.com/automerge/automerge/pull/181 which I think would elevate some of the pain, but having those methods available would be still nice. I'll write some of my thoughts on that in a separate comment.

Gozala:
I do find the fact that change handler gets a mutable ref through proxy quite confusing. It is also worth considering that Proxies are fairly slow and JS engines can't really optimize them as well. With that I would like to propose alternative approach.

[...]

PR #181, opened 2019-07-14
Allow Text instatiation with preset value

### LINKED ISSUE #170: Init Automerge.Text with a String/Array of characters
I would like to propose the following:

```javascript
Automerge.change(savedDoc, doc => {
   doc.someText = new Automerge.Text("Initial Content");
});

``` 
or

```javascript
Automerge.change(savedDoc, doc => {
   doc.someText = new Automerge.Text("I", "n", "i");
});

```
Would be even more powerful if #166 would work

Gozala:
Attempt to fix #170

I'm not super happy about how code turned out as `elems` / `maxElem` are being mutated afterwards, but I also could not figure any better option. I have considered using separate field instead but that had it's own drawbacks, specifically `new Text("hello")` would have appeared as empty text.

### DIFF
--- a/frontend/apply_patch.js
+++ b/frontend/apply_patch.js
@@ -1,6 +1,6 @@
 const { ROOT_ID, isObject, copyObject, parseElemId } = require('../src/common')
 const { OPTIONS, OBJECT_ID, CONFLICTS, ELEM_IDS, MAX_ELEM } = require('./constants')
-const { Text, makeTextInstance } = require('./text')
+const { Text, instantiateText } = require('./text')
 const { Table, instantiateTable } = require('./table')
 const { Counter } = require('./counter')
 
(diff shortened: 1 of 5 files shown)


jeffpeterson:
This would be useful to go along with `Automerge.from`:
`Automerge.from({text: new Automerge.Text("foobar")})`

We run into this exact issue in pushpin.

ept:
Thanks @Gozala! I have tweaked it a bit, and also added the functionality requested in #180. With this PR the following things should now work:

1. `new Automerge.Text('initial text')` or `new Automerge.Text(['i', 'n', 'i', 't'])` to set the initial contents of a Text object
2. `Automerge.from({text: new Automerge.Text('initial value')})`
3. Calling `.insertAt()`, `.deleteAt()` and `.set()` methods on a text object before it is assigned to a document:
    ```js
    Automerge.change(state, doc => {
      const text = new Automerge.Text()
      text.insertAt(0, 'h', 'e', 'l', 'l', 'o')
      doc.text = text
    })
    ```
4. Calling the same methods on a text object after it is assigned to a document, but without going through the proxying flow:
    ```js
    Automerge.change(state, doc => {
      const text = new Automerge.Text()
      doc.text = text
      text.insertAt(0, 'h', 'e', 'l', 'l', 'o') // can call .insertAt on either `text` or `doc.text`
    })
    ```

Gozala:
Thank you! This makes it a lot more pleasant to use

PR #183, opened 2019-07-23
Add initialization options parameter to `automerge.from`

HerbCaudill:
Currently you can either call init with no arguments,
```js
const foo = automerge.init()
```
or you can pass some options around how `actorId` is set:
```js
const foo = automerge.init(actorId)
```
or
```
const foo = automerge.init({actorId, deferActorId})
```

This PR adds the (optional) options parameter to `automerge.from`, so you can do this:
```js
const foo = automerge.from(initialState, actorId)
```
or
```js
const foo = automerge.from(initialState, {actorId, deferActorId})
```

Still to do: 
- [x] add test coverage

### DIFF
--- a/@types/automerge/index.d.ts
+++ b/@types/automerge/index.d.ts
@@ -15,9 +15,13 @@ declare module 'automerge' {
 
   // Automerge.* functions
 
-  function init<T>(actorId?: string): Doc<T>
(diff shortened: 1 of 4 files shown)


ept:
This looks good, thank you @HerbCaudill!

PR #184, opened 2019-07-23
Adjustments to type definitions

HerbCaudill:
This PR fixes ~three~ two minor errors or omissions in the type definitions:

- `DocSet<T>` was missing `docIds` property
- `DocSet.getDoc` return value is `Doc<T>,` not `T` 
- ~Clock is a Map, not an object~

~Still to do:~
- [x] Add test coverage

### DIFF
--- a/@types/automerge/index.d.ts
+++ b/@types/automerge/index.d.ts
@@ -107,7 +107,7 @@ declare module 'automerge' {
   class DocSet<T> {
     constructor()
     applyChanges(docId: string, changes: Change[]): T
-    getDoc(docId: string): T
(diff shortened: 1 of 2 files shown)


HerbCaudill:
- I reverted the `Clock` type change - internal to the `Connection` class, a clock is a `Map`, but elsewhere they're plain objects. 
- I've added a test covering the `docIds` property. 
- We're not currently differentiating between `Doc<T>` and `<T>` in tests, so there's no new test for that change. 

@ept - This is ready for you to take a look. Thanks!

ept:
Merged, thanks @HerbCaudill!

PR #185, opened 2019-07-26
refactor: Remove unused `local` list

edahlseng:
This pull request removes the `local` list from the backend. As far as I can tell, this isn't actually used anymore (I searched the codebase for `'local'`, and this was the only occurrence). If I've missed something, please help me understand what this does!


ept:
I think you're right, it's no longer used. Thanks!

PR #186, opened 2019-07-27
docs: Add local state overview

edahlseng:
This pull request updates the INTERNALS documentation, adding an overview of how the backend actually stores its state. There are a few specific outstanding questions that I still have; I'll call those out inline.


ept:
Thanks @edahlseng!

PR #199, opened 2019-07-31
.toString() of Text should return a string content

Gozala:
I find it really confusing that `new Automerge.Text("hello").toString()` return `"h,e,l,l,o"` instead of `"hello"`. I assume this was not deliberate choice and this pull request changes to do later instead.


ept:
Agree, this is an improvement. Thanks @Gozala!

PR #201, opened 2019-08-06
Add tests for concurrent deletion

HerbCaudill:
Just added a couple of (passing) tests to satisfy curiosity about how concurrent deletes are handled.


ept:
Good, thanks @HerbCaudill!

Issue #202, opened 2019-08-06
`Maximum call stack size exceeded` error when initializing a document with a given size


HerbCaudill:
```js
      describe.only('instantiation from large objects', () => {
        const largeObject = N => {
          const record = { first: 'Martin', last: 'Kleppmann' }
          const list = []
          const map = {}
          for (let id = 0; id < N; id++) {
            const r = { id, ...record }
            list.push(id)
            map[id] = r
          }
          return { list, map }
        }

        it('should handle 20000 records', () => 
          assert.doesNotThrow(() => Automerge.from(largeObject(20000)))
        ).timeout(20000)

        it('should handle 20895 records', () => 
          assert.doesNotThrow(() => Automerge.from(largeObject(20895)))
        ).timeout(20000)

      })
``` 

The first test (with 20000 records) passes, but the second (with 20895 records) throws a stack overflow. 
```

  Automerge
    sequential use:
      initialization
        instantiation from large objects
          √ should handle 20000 records (5806ms)
          1) should handle 20895 records


  1 passing (11s)
  1 failing

  1) Automerge
       sequential use:
         initialization
           instantiation from large objects
             should handle 20895 records:
     AssertionError [ERR_ASSERTION]: Got unwanted exception.
Actual message: "Maximum call stack size exceeded"
      at Context.<anonymous> (test\test.js:74:18)
      at processImmediate (internal/timers.js:439:21)
```

HerbCaudill:
Probably same issue as #137

ept:
Fixed by avoiding calling `Array.push` with a very long list of arguments: 5ca49f6b99f15709959103fa3cae6789b8945ebf

PR #203, opened 2019-08-06
Fix stack overflow when initializing with large documents

HerbCaudill:
Opening draft PR with a failing test in anticipation of fixing #202


HerbCaudill:
Fixed by 5ca49f6

PR #206, opened 2019-08-14
Fix trap error by setting writable property descriptor to true

### LINKED ISSUE #205: TypeError: 'getOwnPropertyDescriptor' on proxy in chrome 76+
Hey, first of all I'd like to thank you for all the hard work on this library. I only recently started using automerge and love the paper + library 😍 	

After chrome was updated last week, I experienced some issues with proxy lists.

I tried replicating the issue in a codesandbox but then realized that there are browser tests validating the exact behavior I'm having trouble with.

Run `yarn build && yarn browsertest` to see following error

```
Chrome 76.0.3809 (Windows 10.0.0) Automerge proxy API list object should support Object.keys() FAILED
  TypeError: 'getOwnPropertyDescriptor' on proxy: trap reported non-configurable and writable for property 'length' which is non-configurable, non-writable in the proxy target
    at Function.keys (<anonymous>)
```

I'm running following software
```
Chrome: 76.0.3809.100
Windows 10: OS Version 1809 (Build 17763.615)
```

ReasonableDeveloper:
Attempts to fix #205

### DIFF
--- a/frontend/proxies.js
+++ b/frontend/proxies.js
@@ -170,7 +170,7 @@ const ListHandler = {
   },
 
   getOwnPropertyDescriptor (target, key) {
-    if (key === 'length') return {}
+    if (key === 'length') return {writable: true}
     if (key === OBJECT_ID) return {configurable: false, enumerable: false}
 
     const [context, objectId] = target


ept:
Thank you, this looks good!

PR #208, opened 2019-08-15
Add initialization tests for array, primitives

HerbCaudill:
1. Moves all initialization tests to an 'initialization' `describe` block directly under 'Automerge' 
2. Adds tests for passing different types of values to `Automerge.from`: an array, a string, a number, and a boolean.


ept:
Hi @HerbCaudill, I'm happy to add tests, but could we avoid moving tests around for now if it's not strictly necessary? The reason is that I have a big refactoring in progress (on the `performance` branch), and the merge conflicts with that branch are going to be a headache.

PR #210, opened 2019-08-19
Add DocSet.removeDoc

brentkeller:
Adds functionality to remove a doc from a `DocSet`. I'm not totally sure how the DocSet handlers work so I wasn't sure if it was appropriate to call the handlers with the deleted document's ID and `undefined`. Should that be added as well?


ept:
This looks good to me, thanks @brentkeller!

PR #215, opened 2019-10-19
Support freezing of opaque string types

jeffpeterson:
We often use `type FooUrl = string & { __fooUrl: void }` to give types to different kinds of strings.
The current `Freeze<T>` type breaks these. This PR is a small fix that we've been using for awhile that has `Freeze<T>` treat these types like normal strings.


ept:
Thank you @jeffpeterson!

PR #218, opened 2019-11-03
Add Perge to list of Send/Receive libraries in README

sammccord:
I've been messing around with Automerge for a while now and I'm big fan, so I took a stab at writing a library to integrate PeerJS. Since MPL was a little more than I needed, and PeerJS makes working with WebRTC DataChannels dead simple, I figured this combo would be super pleasant.

Check out the project and feel free to reach out with any feedback!

### DIFF
--- a/README.md
+++ b/README.md
@@ -357,6 +357,8 @@ options, with more under development:
 - [Hypermerge](https://github.com/automerge/hypermerge) is a peer-to-peer networking layer that
   combines Automerge with [Hypercore](https://github.com/mafintosh/hypercore), part of the
   [Dat project](https://datproject.org/).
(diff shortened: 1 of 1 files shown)


ept:
Hi @sammccord, this looks great, thank you! I've added the link to the README.

Can I just make one suggestion: in your example, you have a counter that you increment using `++`. This type of counter will not work correctly when multiple users concurrently increment it, as explained in [this section of the README](https://github.com/automerge/automerge#counter). Can I suggest using `Automerge.Counter` instead?

sammccord:
> Can I suggest using `Automerge.Counter` instead?

Whoops, fixed!

Issue #230, opened 2020-02-07
Text.deleteAt deletes a character with 0 as delete count


skokenes:
Hi, I think there's slightly odd behavior with the Text.deleteAt function. If you pass in 0 as the argument for deletion count, it still deletes a character. For example:
```javascript
const t = new Automerge.Text("foo");
t.deleteAt(0,0);
t.toString(); // will yield "oo"
```

I believe this happens because of the following code in the `deleteAt` method here https://github.com/automerge/automerge/blob/master/frontend/text.js#L109
```javascript
this.elems.splice(index, numDelete || 1)
```

If you pass numDelete as `0`, that will evaluate as false and therefore `1` will be inserted instead.

The use case for passing 0 is when autogenerating Automerge changes from another source, like Slate transforms which sometimes generate text removal calls of 0 length. The workaround is easy: just filter those calls out. 

But, it is still unexpected to see this behavior when using Automerge, even if its easily avoided. I'd expect `deleteAt` to behave similarly to an Array.splice() call, which won't remove anything if the number provided is `0`.

pvh:
Great bug report! I agree this is a defect and if you want to send a little PR for that with a test I'll merge it today.

PR #231, opened 2020-02-07
fix deleteAt bug with input 0

skokenes:
Fixes https://github.com/automerge/automerge/issues/230


pvh:
Excellent, thank you!

PR #235, opened 2020-03-09
Make the README caveats a bit less scary

jjlee:
Trying to avoid needlessly scaring potential users like me :-)

Hope my wording is reasonable and accurate, but of course you might want to express it differently, this is just a suggestion...

### DIFF
--- a/README.md
+++ b/README.md
@@ -659,9 +659,15 @@ possible to have row objects that don't have values for all columns (e.g. in the
 
 The project currently has a number of limitations that you should be aware of:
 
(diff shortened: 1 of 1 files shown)


ept:
Thank you @jjlee for these suggestions. I have rephrased the caveats a bit differently in 809887c7e2928639d85490fff82bbbeef0bea707. I ended up taking out the bullet point about integrity checking because AFAIK no other CRDT library does this, and so there probably isn't an expectation around this.

PR #236, opened 2020-03-11
Simpler `Table` type

HerbCaudill:
### Summary

This PR removes an undocumented API for adding rows to a `Table` by passing an array of values, and simplifies the type definitions for tables accordingly. 

### Problem

Using the Table CRDT, you can add a row by passing an object:

```js
  const ddia = doc.publications.add({
    type: 'book',
    authors: [martinID],
    title: 'Designing Data-Intensive Applications',
    publisher: "O'Reilly Media",
    year: 2017,
  })
```

Or, you can pass an array of values:

```js
const ddia = doc.publications.add([
    'book', 
    [martinID],  
    'Designing Data-Intensive Applications', 
    "O'Reilly Media", 
    2017
])
```

Here Automerge automagically figures out which value goes in which column, based on the order of the columns provided when initializing the table. 

This array-of-values API is undocumented and, I'm guessing, unused. Supporting it in Typescript makes the type definition for a table fairly awkward, because in addition to defining a type for rows, you have to redundantly list the fields in order:

```ts
interface Book {
    authors: string | string[]
    title: string
    isbn?: string
}

interface BookDb {
    books: Automerge.Table<Book, ['authors', 'title', 'isbn']> // <-- 
    // Q: Why do I have to list these fields again? 
    // A: So that Typescript can enforce column types if you ever decide 
    // to use the undocumented API to add a row as an array of values. 🙄
}

let s1 = Automerge.init<BookDb>()
s1 = Automerge.change(s1, doc => {
    doc.books = new Automerge.Table(['authors', 'title', 'isbn'])
    id = doc.books.add(DDIA)
})
```

This has been [confusing](https://automerge.slack.com/archives/C61RJCM9S/p1580753919087700?thread_ts=1580644956.082000&cid=C61RJCM9S) for developers trying to use Automerge with TypeScript. 

### Proposed solution

- Remove the alternative API, so that table rows have to be added in object form. 
- Change the way tables are defined in TypeScript, so you only need to pass the underlying type and not the type *and* an ordered array of column names.

### Questions

This is a breaking change, in two ways:

- If anyone was actually using the array API, they will now (silently) get unexpected results. Should we throw an error instead if an array is passed to `Table.add()`?

- Anyone who defined has previously implemented a `Table` in TypeScript will now get a compilation error:

  ```ts
  // ❌ Array of column names used to be required, now won't compile
  books: Automerge.Table<Book, ['authors', 'title', 'isbn']> 
  // 👍
  books: Automerge.Table<Book> 
  ```

  Should we continue to support the old syntax instead, and just mark it as deprecated in the `Table` type definition?

### DIFF
--- a/@types/automerge/index.d.ts
+++ b/@types/automerge/index.d.ts
@@ -55,9 +55,9 @@ declare module 'automerge' {
 
   // custom CRDT types
 
-  class Table<T, KeyOrder extends Array<keyof T>> extends Array<T> {
-    constructor(columns: KeyArray<T, KeyOrder>)
-    add(item: T | TupleFromInterface<T, KeyOrder>): UUID
+  class Table<T> {
+    constructor(columns: (keyof T)[])
+    add(item: T): UUID
     byId(id: UUID): T
     columns: string[]
     count: number
@@ -96,7 +96,7 @@ declare module 'automerge' {
 
   // Readonly variants
 
-  type ReadonlyTable<T, KeyOrder extends Array<keyof T>> = ReadonlyArray<T> & Table<T, KeyOrder>
+  type ReadonlyTable<T> = ReadonlyArray<T> & Table<T>
   type ReadonlyList<T> = ReadonlyArray<T> & List<T>
   type ReadonlyText = ReadonlyList<string> & Text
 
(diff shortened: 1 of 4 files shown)


ept:
This is a good simplification, thank you @HerbCaudill. I have made it throw an exception if an array is passed to `Table.add()`. I don't think it's necessary to go through deprecation for this, since it wasn't documented and since Automerge is still in version 0.* (so SemVer allows us to make breaking changes).

I wonder if we should go even further and remove the list of columns passed to the Table constructor, since those columns now aren't actually used for anything. What do you think?

HerbCaudill:
> I wonder if we should go even further and remove the list of columns passed to the Table constructor, since those columns now aren't actually used for anything. What do you think?

Yeah, I had the same thought but wasn't sure if I was missing somewhere they were used.

ept:
Okay, I've put up PR #241 to completely remove the columns list.

PR #238, opened 2020-03-17
Allow non-string children in Text objects

### LINKED ISSUE #194: Inserting non text characters into Automerge.Text
This is related to #193 as I attempt to model rich text as series of tokens of characters and markers:

```js
{format:{italic:true}}Hello {format:{bold:true, italic:true}}beautiful{format:null} world\n
```

As I was trying to figure out a best approach to this I noticed that `Automerge.Text` would happily insert objects e.g. `text.insertAt(0, { bold: true })` which than would appear as `objectId`s inside the text. ([related thread on slack](https://automerge.slack.com/archives/C61RJCM9S/p1563228180398800)) It's also worth mentioning that `Automerge.getObjectById(doc, objectId)` would actually return inserted object.

I'm not sure what is the intended behavior here, but I suspect that `text.join('')` including `objectId` is highly unlikely. It seemed like a convenient accident for my use case, but I'm suspicious there might be deeper consequences so I chose to use list (or should I say array) of tokens instead. Either way I would like to propose:

1. To document how `Automerge.Text` outperforms arrays (as suggested by the readme).
2. Either throw if non-text is being inserted or omit `objectId`s in `.join()` and document non textual inserts bit more.

nornagon:
This builds on work from @pvh and @ept to allow non-string objects to be present in `Automerge.Text`. While the experiments with representing rich-text formatting by "format markers" proved unsuccessful, I think this is still useful for representing "embedded objects" within text.

Closes #194

### DIFF
--- a/@types/automerge/index.d.ts
+++ b/@types/automerge/index.d.ts
@@ -76,7 +76,7 @@ declare module 'automerge' {
     constructor(objectId?: UUID, elems?: string[], maxElem?: number)
     get(index: number): string
     getElemId(index: number): string
-    toSpans(): (string | any)[]
+    toSpans<T>(): (string | T)[]
   }
 
   // Note that until https://github.com/Microsoft/TypeScript/issues/2361 is addressed, we
--- a/@types/automerge/index.d.ts
+++ b/@types/automerge/index.d.ts
@@ -76,6 +76,7 @@ declare module 'automerge' {
     constructor(objectId?: UUID, elems?: string[], maxElem?: number)
     get(index: number): string
     getElemId(index: number): string
+    toSpans(): (string | any)[]
   }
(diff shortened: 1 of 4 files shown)


nornagon:
I think this closes #194?

ept:
Merged, thank you!

ept:
Note for posterity: this PR also includes @pvh's initial implementation of the OAFS model in `test/text_test.js`. For now that seems as good a place as any to keep it, but if we want to make it available as part of the Automerge API we'll have to move it into the main library code.

nornagon:
@ept I don't think this includes any OAFS stuff, though it does contain a pre-OAFS attempt to model rich text using inline markers in the Text object.

Issue #239, opened 2020-03-17
Proposal for a new Automerge.TextSpan type


nornagon:
In order to support the Ordered Array of Formatting Spans ([OAFS][]) model for Rich Text, it's necessary to maintain a list of references into an Automerge.Text object, along with the formatting attributes to be applied to the given range. For illustration, consider represent the string "Gandalf the Grey", where the string "Gandalf" is bold:

```json
{
  "text": "Gandalf the Grey",
  "attributes": [
    { "from": 0, "to": 7, "attributes": { "bold": true } }
  ]
}
```

However, in order to make sure the references remain consistent across operations, instead of using index references, one should store the ID of the referenced character. IDs for individual characters are not currently exposed by Automerge, and further, IDs would need to be translated back to string offsets in order to be meaningful to a client.

Spans would also be useful for representing and transforming user cursors/selections.

## API

Creating a TextSpan would be done from the corresponding Text object:

```js
const text = new Automerge.Text('Gandalf the Grey')
const span = text.getSpan(0, 7)
```

Spans are Automerge objects, so they can be stored in a document:

```js
const doc = Automerge.from({
  text: new Automerge.Text('Gandalf the Grey'),
  attributes: []
})
doc.change(d => {
  const span = d.text.getSpan(0, 7)
  d.attributes.push({ span, attributes: { bold: true } })
})
```

Spans can be inspected, and have "start" and "end" properties that are the calculated offsets into the Text object which they reference:

```js
doc.attributes[0].start === 0
doc.attributes[0].end === 7
```

When the associated Text has characters inserted or removed, Span objects will stay up-to-date:

```js
const text = new Automerge.Text('Gandalf the Grey')
const doc = Automerge.from({
  text,
  span: text.span(0, 7)
})

const doc2 = doc.change(d => {
  d.text.insertAt(3, 'non')
})

doc2.span.end === 10
```

[OAFS]: https://github.com/automerge/automerge/issues/193#issuecomment-544646991

nornagon:
Hm, one potential complication with implementing this by storing IDs to the beginning/end of the span is that the IDs would need to be updated when the characters to which they refer are deleted.

For instance:

```
Gandalf the Grey
    ^              here's my pointer
  |___|            delete these 5 characters
Ga the Grey        new string
 ^                 pointer should end up here
```

ept:
Thank you for the proposal. We definitely need something along these lines, and I hope we can work out a good API together.

[...]

PR #240, opened 2020-03-18
allow getting index for id from text

nornagon:
This adds a new API to Automerge.Text to get the index of a character from its ID. It is the inverse operation of `getElemId`.

The imagined use-case here is for storing offsets into text, related to #239

### DIFF
--- a/frontend/text.js
+++ b/frontend/text.js
@@ -27,6 +27,10 @@ class Text {
     return this.elems[index].elemId
   }
 
+  getIndexForId (id) {
+    return this.elems.findIndex(e => e.elemId === id)
+  }
+
   [Symbol.iterator] () {
     let elems = this.elems, index = -1
     return {
--- a/test/text_test.js
+++ b/test/text_test.js
@@ -170,4 +170,46 @@ describe('Automerge.Text', () => {
       assert.strictEqual(s1.text.toString(), 'Init')
     })
   })
+
+  describe('element ids', () => {
+    it('should return ids for every element in the text', () => {
+      const s1 = Automerge.from({text: new Automerge.Text('init')})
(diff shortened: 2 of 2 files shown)


nornagon:
Hm, this doesn't account for tombstoned IDs. Closing while I consider a different approach.

ept:
Something along these lines would be useful, but I am not yet sure as to the best way of doing it. On the `performance` branch I've been working on I have actually removed the `elemIds` array from Automerge.Text, because I would like to keep the frontend as simple and lean as possible, and maintain this metadata in the backend instead. For users who are running everything on one thread, this is not a problem, but for users who are running frontend and backend on separate threads, it will no longer be possible for the Automerge.Text object to look up elemIds.

Perhaps a better approach is to use marker objects in text, like in your other PR #238 (which I am still reviewing)?

nornagon:
@ept hm, using marker objects for spans seems intuitively wrong for a few reasons, but maybe my intuition is wrong. Here's what comes up for me when I think about that:

- One end of a range could be deleted, but not the other end.
- Cursor positions would have to be "updated" when inserting a range, despite no visible cursor change.
- Insertion order isn't guaranteed to be sensible, i.e. if I create a range and you create a range over the same text, whose range comes "first"?

The marker objects in #238 are best suited, I think, for non-text objects inline with the text, for instance, images or "tags".

I still think the best API here is probably one that allows the user to specify offsets into the string, and which under the hood is translated between offsets on the frontend and IDs on the backend.

ept:
Okay, that makes sense. There should be a way of referring to spans/cursor positions without modifying the sequence itself. Let's continue the discussion about that on #239.

PR #241, opened 2020-03-18
Remove the list of columns from Automerge.Table

ept:
Since #236 the Automerge.Table type no longer allows rows to be added as an array of values, which are then mapped to column names. Now that this feature is removed, nothing in Automerge is actually using the list of columns that is currently required by the Automerge.Table constructor. The columns are still saved, but since they do not enforce any schema, they are purely advisory.

Hopefully one day we can have proper schema support in Automerge, but the current half-hearted implementation is not really helping us get there. Therefore I think it is best to just remove this list of columns feature.

Any thoughts?

### DIFF
--- a/@types/automerge/index.d.ts
+++ b/@types/automerge/index.d.ts
@@ -56,16 +56,14 @@ declare module 'automerge' {
   // custom CRDT types
 
   class Table<T> {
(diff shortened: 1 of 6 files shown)


ept:
Merged this as there has been no objection.

PR #242, opened 2020-03-23
Make table row's primary key available as `id` property

ept:
While we're making API changes to `Automerge.Table` anyway (#236, #241), here's another change I'd like to make.

Currently, if you have a row object from an Automerge.Table, you can get its primary key with `Automerge.getObjectId(row)`. This API is not very discoverable; users who don't know about this API might be tempted to generate their own IDs for table rows, which would miss the entire point of Automerge.Table.

This patch makes the primary key more visible by making it available as `row.id` instead. When a new row is added, we check that the row object doesn't already have an `id` property. We also ensure that the `id` property cannot be modified.

For example:

```js
let bookId
let database = Automerge.change(Automerge.init(), doc => {
  doc.books = new Automerge.Table()
  // the add() method returns the primary key of the new row
  bookId = doc.books.add({
    author: 'Martin Kleppmann',
    title: 'Designing Data-Intensive Applications',
    year: 2017
  })
})

database.books.byId(bookId)
// Returns {
//   id: bookId,  <--- new property!
//   author: 'Martin Kleppmann',
//   title: 'Designing Data-Intensive Applications',
//   year: 2017
// }
```

(Besides API usability, there is also a deeper reason for making this change: on the `performance` branch, objectIds are backend-generated Lamport timestamps rather than UUIDs to enable better compression; since `Table.add` should synchronously return the primary key of the new row, it must use a different ID from the objectId. Putting the row's primary key in a separate property reinforces that distinction.)

### DIFF
--- a/@types/automerge/index.d.ts
+++ b/@types/automerge/index.d.ts
@@ -55,14 +55,18 @@ declare module 'automerge' {
 
   // custom CRDT types
 
+  class TableRow {
+    readonly id: UUID
+  }
+
   class Table<T> {
     constructor()
     add(item: T): UUID
-    byId(id: UUID): T
+    byId(id: UUID): T & TableRow
     count: number
     ids: UUID[]
     remove(id: UUID): void
-    rows(): T[]
+    rows(): (T & TableRow)[]
     set(id: UUID, value: T): void
   }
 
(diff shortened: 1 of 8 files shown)


pvh:
I suspect there will be a lot of extant data out there that already uses the name "id". Should we consider an alternate name like objectId? Could we also, perhaps, extend this (or something like this) to arrays?

ept:
> Should we consider an alternate name like objectId?

Could do, although my preference would be to make the API as straightforward as possible now, so that it is nice in the long term (hopefully the number of users we have now is tiny in comparison to the number we will have in the future). I'm not keen on `objectId` since the performance branch is breaking the link between primary keys in a table and object IDs. But we could use e.g. `_id` if we want to pick a more obscure name.

> Could we also, perhaps, extend this (or something like this) to arrays?

Not sure what you mean with this. Accessing list element IDs? On the performance branch I have removed all knowledge of element IDs from the frontend, so this would not be possible.

PR #249, opened 2020-04-14
Update table class type to reflect property getter

lauritzsh:
The `rows` here is a property getter and not a function that can be called from what I can see [here](https://github.com/automerge/automerge/blob/14409f01adae1778c76ca13e7432f395fea4db54/frontend/table.js#L60-L66). As it is right now I can't use it since TypeScript think it's a function and calling gives an error.


ept:
Well spotted, thank you for the correction!

PR #252, opened 2020-04-29
Fix type signatures for WatchableDoc#get and WatchableDoc#set

vincentcapicotto:
This change reflects the internal implementation, it's also consistent with the `DocSet#set` and `DocSet#get` signatures


ept:
Merged, thanks @vincentcapicotto!

Issue #200, opened 2019-08-06
Using hash chaining to encode the dependency graph


edahlseng:
While looking through previous issues and pull requests, I’ve noticed a few mentions of a desire to move from vector clocks towards hash chaining for encoding the dependency graph (#27, #28, #31, #36). After reading through these issues and pull requests, there are a number of questions that come to mind:
* Is this still something that is desired? (If not, then the rest of the questions below become irrelevant!)
* Would it be okay if I started working towards adding this functionality?
    * It seems like adding hash-based integrity checking (#27) would be a good place to start, with full hash chaining to follow. Does this make sense?
    * Are there any backwards-compatibility concerns that would need to be kept in mind?
* What are the expected benefits of using hash chaining over vector clocks? A few potential benefits that jump to mind:
    * Dependency encoding becomes a fixed size for each change, rather than the `O(# of peers)` size that comes with vector clocks (though that’s only observed in the worst case, given the recent change to remove transitive dependencies from changes)
    * I _think_ we wouldn’t need to store the `states` object any more, instead we would only need to store a collection of each hash and its parent(s)
    * I _think_ we wouldn’t need to store `clock` or `deps` objects any more, instead we would only need to store a set of the most recently observed concurrent hashes
    * Any more?
* What are the expected drawbacks of using hash chaining over vector clocks?
  * One of the issues (#31) mentions that we would need to use a multi-round protocol to determine the latest common ancestor between peers
  * Any others?

ept:
Hi @edahlseng, I think this is still desirable, but I am not yet sure how exactly this feature should look like, and so it may take a few iterations before we are happy with it. You're welcome to explore the topic; I just say that to warn that the result might not be a quick "looks good to me, merge the PR" but might need some more deliberation.

In terms of efficiency, I think that hash chaining won't make a big difference. As you say, the vector clocks in changes already encode just the minimal dependencies (excluding transitive dependencies), so the size of encoding the dependencies with hashes would be roughly the same. And in terms of the data structures that need to be maintained, yes, we can remove some, but we also need to add a big map that allows changes to be looked up by their hash, which will probably have about the same size.

[...]

Issue #89, opened 2018-05-09
Performance/scalability?


aral:
Hey there,

First off, thank you for sharing your paper (A Conflict-Free Replicated JSON Datatype) and this library with the world.

I started playing with automerge this afternoon and wanted to run some basic tests to see if I could use it for the federated personal web site system I’m working on, as I want it to work offline and I’d rather not privilege servers over clients (the goal is, ideally, for this to be a stepping stone towards a p2p world, which I believe is what we’re all working for) :)

So to cut to the chase, I wanted to share my very unsophisticated findings and to also inquire about whether, in the words of Steve Jobs, I’m “holding it wrong” :)

## Multiple inserts

**Code:** https://source.ind.ie/snippets/80

I first tried pushing incremental numbers to an array in a document via the `change()` method. The timings (on my MacBook) ranged from 6ms for 1 push to 43.75 seconds for 10,000. 100 pushes took 175ms and 1,000 took 1.345 seconds.

<img width="675" alt="screen shot 2018-05-09 at 21 58 41" src="https://user-images.githubusercontent.com/41798/39839659-c901c828-53d4-11e8-9e90-ae80ec0e0089.png">

The storage requirements (tested very roughly using `util.inspect` to flatten the object, including all hidden fields) also seem to follow a similar curve. A single insert took up 3,557 bytes (size of original object: 18 bytes), 10: ~17KB, 100: ~160KB, 1,000: ~1.6MB, and the array with 10,000 numbers took up ~ 16MB (size of original object: ~80 bytes). Again, my testing methodology doesn’t necessarily reflect the actual amount of space these objects would take up either in memory or on the file system but, unless I’ve done something really daft (which is always a possibility), it does look like the system would result in a sluggish interface on operations on an array with ~100 items or so.

<img width="707" alt="screen shot 2018-05-09 at 21 58 47" src="https://user-images.githubusercontent.com/41798/39839685-df8829b6-53d4-11e8-9262-a46bc426d5ac.png">

## Single insert

Then I thought maybe I am holding this wrong and it is meant to be used with batch inserts.

**Code:** https://source.ind.ie/snippets/81

So instead of testing, say, 10,000 pushes to an array, I wanted to test a single change to the object where an array with 10,000 items is added.

The results I got mirror those of the first scenario.

The timings seem to follow a similar curve at first but the results I got for the array with 10,000 items was ~3x slower at ~115 seconds vs ~43 seconds using the first technique.

[...]

Issue #27, opened 2017-06-22
Hash-based integrity checking of operation sets


ept:
At the moment, tesseract has no integrity checks. For example, a buggy node could send two different operations with the same sequence number to different peers, and they may never notice that they have diverged.

A simple approach for fixing this would be to take a leaf out of git's model: each operation has a cryptographic hash that covers the contents of the operation, the author, and the hashes of the prior operations on which it depends. The recipient must check that the hashes match up. Such hash-chaining alone is not sufficient to prevent disruption by actively malicious peers, but it would catch a lot of accidental bugs, and it would pave the way for a protocol that is robust against malicious peers (e.g. using digital signatures).

ept:
Hash chaining of Automerge changes is implemented as part of the new binary format #253. Implications for networking layers are discussed in #290. I'm going to close this issue and leave further discussion on those issues.

PR #253, opened 2020-05-01
Preview: Automerge binary data format

ept:
As we all know, Automerge's current performance on large documents is [terrible](https://github.com/dmonad/crdt-benchmarks) (#89) — loading them is very slow, the data files on disk are huge, and they use huge amounts of memory and network bandwidth. Improving this situation is one of my top priorities. However, most of the low-hanging fruit had already been picked, and a more fundamental rethink of Automerge's data structures was needed.

It turns out that CRDTs are very easy to implement badly, but actually quite difficult to make fast and efficient. A lot of this is due to the metadata overhead: for example, in a text document, every single character needs a unique identifier. Various schemes have been designed to reduce this metadata cost, but some of them [behave badly under concurrent insertion](https://martin.kleppmann.com/papers/interleaving-papoc19.pdf). Others [rely on periodic cleanup operations](https://hal.inria.fr/hal-02526724/document), but the cleanup depends on communicating with all nodes, and it stops if some of the nodes are offline for a long time.

Almost a year ago I first [wrote up a proposal](https://github.com/automerge/automerge-perf/blob/master/columnar/README.md) for a compressed binary encoding format for Automerge, taking a new approach that I've not yet seen in the CRDT research literature. Borrowing ideas from [column-oriented database systems](http://www.cs.umd.edu/~abadi/papers/abadi-column-stores.pdf), this is in the first instance a way of saving disk space and network bandwidth, but it also paves the way towards some big potential performance improvements (and it doesn't suffer from the aforementioned problems).

I have been working on this project on and off since, and while it's not yet finished, I wanted to share an update on where we're at and where we're heading. I am excited about the progress so far and I'm pretty sure this is the right path for the future.

This work has been happening on the [`performance`](https://github.com/automerge/automerge/compare/performance) branch. My intention is to stabilise the features and data formats on this branch over the coming weeks, and to then turn it into a release candidate for ✨**Automerge 1.0**✨. There are a bunch of compatibility-breaking changes on this branch, both in the APIs and the data formats, and I want to wrap up all of these breaking changes into a single bump of the major version number, so as to minimise the number of further breaking changes in the future. There will be a migration tool for Automerge 0.* users to convert their existing documents to the Automerge 1.0 format, and the new data formats are carefully designed with extensibility in mind, allowing us to add new features to Automerge while retaining interoperability between clients running different versions of Automerge.

`Automerge.getChanges()` and `Automerge.save()` now return `Uint8Array` objects using the binary data format. The Automerge frontend APIs are largely unchanged (except for a few tweaks as documented in the [CHANGELOG](https://github.com/automerge/automerge/blob/performance/CHANGELOG.md)). The communication protocol between frontend and backend [has changed a lot](https://github.com/automerge/automerge/blob/performance/BINARY_FORMAT.md) (fixing some historical design mistakes, and allowing the frontend to be simplified, which should make it easier to create frontends in other languages). 

There are also some changes of the "might as well also do this now" sort:

* Every change now has a millisecond-resolution timestamp of when it was made. This timestamp has no bearing on the conflict resolution; it exists only to enable better inspection of the editing history (allowing you to see *when* a change was made). I'm not sure why we didn't add a timestamp from the beginning — it seems like quite an obvious omission in retrospect.
* Every change can be identified by the SHA-256 hash of its binary encoding, and this hash is now used instead of actorId+sequence number (vector clocks) to express dependencies between changes — much like in Git or blockchains (see discussion in #27 and #200). ActorIds still exist, because using hashes for everything totally destroys compression (see below).

The switch to using hash chaining will have a bearing on network protocols that sync up Automerge replicas, such as Automerge.Connection, which currently relies on vector clocks. This is something I would like to discuss with the community. By the way, I have split Automerge.Connection into a [separate repository](https://github.com/automerge/automerge-connection), because I think we might want to evolve it and assign it version numbers independently from the Automerge core.

Anyway, time to report some numbers on how the binary format compares to Automerge's existing JSON data format. First of all, let me emphasise that I have not yet solved the performance problems on this branch (despite the branch name "performance") — so far I have primarily worked on making the binary format compact. Therefore the analysis below focusses mostly on encoded data sizes. Also, I will use text editing examples, since text editing tends to accumulate changes very quickly (one change per keystroke), so it's a challenging type of workload. Nevertheless, the binary format should be good for all types of data supported by Automerge.

The binary-encoded length of a typical single-character text insertion is 105 bytes. Sounds like a lot, but only 35 of those bytes are the encoding of the actual operation. The rest is made up of the hash of the previous change (33 bytes), the actorId (17 bytes), the timestamp (6 bytes), a checksum (4 bytes), and miscellaneous other header fields (10 bytes). It might be possible to squeeze that a little further, but with diminishing returns. A single-character deletion takes 109 bytes. There is no point gzipping these short byte sequences, as doing so actually *increases* the size.

|                            | Current JSON format | JSON + gzip       | New binary format | Binary + gzip |
|----------------------------|---------------------|-------------------|-------------------|---------------|
| Single-character insertion | 280 bytes           | 187 bytes         | 105 bytes         | 138 bytes     |
| Single-character deletion  | 183 bytes           | 159 bytes         | 109 bytes         | 141 bytes     |
| 10,000-character insertion | 2,476,915 bytes     | 105,296 bytes     | 10,124 bytes      | 4,380 bytes   |
| 10,000-character deletion  | 1,118,968 bytes     | 28,386 bytes      | 133 bytes         | 151 bytes     |

The benefits of the binary format show up much more starkly when inserting or deleting longer runs of characters in one go. When inserting 10,000 ASCII characters at once into a text document (a medium-sized copy&paste), the JSON format uses about 248 bytes per inserted character, leading to a change about 2.5 MB in size. On the other hand, the binary format contains the raw ASCII text (10,000 bytes) plus a small, near-constant overhead of 124 bytes for the whole change. Although the JSON data compresses well with gzip, with a 24:1 compression ratio, the gzipped JSON is still 10 times the size of the un-gzipped binary data. The binary data can still be gzipped further for an additional 2.3:1 compression.

The situation is even more extreme for a 10,000 character deletion: the binary encoding is almost constant-size, while the JSON format takes over 100 bytes per deleted character. In this example, the JSON encoding is 8,400 times the size of the binary encoding.

So far for the microbenchmarks. A more interesting question is how the encoding fares with more complex editing patterns. For this, I used a [dataset](https://github.com/automerge/automerge-perf) that we captured a few years ago, when we wrote the LaTeX source of an [entire paper](https://arxiv.org/abs/1608.03960) using a homegrown text editor. The result is an editing history containing 332,702 keystrokes (of which 182,315 are single-character insertions; the rest are single-character deletions and cursor movements). The final text file (without any editing history) is 104,852 bytes in size.

I [converted](https://github.com/automerge/automerge-perf/blob/master/edit-history/compress.js) this dataset to the new binary format, with each keystroke as a separate change, and the result is as follows:

|                            | JSON format         | JSON + gzip       | Binary format     | Binary + gzip |
|----------------------------|---------------------|-------------------|-------------------|---------------|
| Paper (individual changes) | 146,406,415 bytes   | 71,347,241 bytes  | 51,620,565 bytes  | N/A           |
| Paper (whole document)     | 146,406,415 bytes   | 6,132,895 bytes   | 1,119,341 bytes   | 664,268 bytes |

The first row shows what happens when we treat each of the 332,702 changes separately: 440 bytes per change in JSON, and 152 bytes per change for the binary encoding — a bit more than the per-change numbers above, but not wildly so. However, treating each change individually drastically limits the compression we can apply. Applying gzip compression to each JSON change individually yields a compression ratio of only 2:1.

It is much more effective to compress the document as a whole. Simply gzipping the JSON change history as a whole yields a 24:1 compression ratio. But the big deal is the whole-document binary encoding, which is about a megabyte (3 bytes per change), and which further gzips to 664 kB. When working with the whole document, binary encoding + gzip is almost 10 times more compact than JSON + gzip!

I should emphasise that in all of these examples, the compression is **lossless**: it's not just a snapshot of the latest state, but it preserves the entire character-by-character editing history. In fact, from the whole-document binary format it is possible to reconstruct the bitwise identical change history that created it, recompute all the SHA-256 hashes for the dependency chains between changes, and end up with exactly the same hashes for the "heads" (in Git terminology). Note that hashes need to be recomputed, not stored: merely storing the hash of each of the 332,702 changes would require about 10MB, ten times the size of the binary-encoded file.

The next interesting question is: what makes up those 1,119,341 bytes of binary data? Can we reduce it further? Well, it breaks down as follows:

* 662,388 bytes for the timestamps on each change. Yes, you read that correctly: two thirds of the file is just timestamps! The reason is: timestamps are millisecond resolution, and for each change we store the number of milliseconds since the previous change. When typing, the time between keystrokes is usually a few hundred milliseconds. The binary format encodes integers between 64 and 8,191 using two bytes, so most changes require two bytes to represent their timestamp delta to the previous change. Two bytes times 332,702 changes costs about 664 kB. We could reduce this to one byte per change if we reduce the timestamp resolution to integer seconds instead of milliseconds.
* 182,315 bytes for the ASCII characters of the 182,315 insertion operations. This is just ASCII text, so it further gzips pretty well. This example document only contains ASCII, but obviously the binary format handles all of Unicode using UTF-8.
* 112,726 bytes for the CRDT metadata on the text object. That is about 0.44 bytes per text editing operation — pretty awesome!
* 124,306 bytes for the cursor movement operations (1.2 bytes per operation).
* 36,954 bytes to store how the operations are grouped into changes.
* 652 bytes for miscellaneous headers.

Comparing the whole-document binary encoding (~1 MB without gzip) to the final text document without any CRDT metadata (~100 kB without gzip), we see there is still a substantial cost. But we do gain a very detailed change history, and of course the CRDT merging ability. And we have made big progress compared to the JSON encoding (two orders of magnitude smaller without gzip, one order of magnitude smaller with gzip). We could compress it by another factor of ~2 by reducing timestamps to 1-second resolution and leaving out the cursor movements (they could be maintained in separate transient storage, and omitted from the persistent document history).

So far all of the compression has been lossless. Of course we can also consider approximate forms of compression, such as combining several changes with similar timestamps into a more coarse-grained change (reducing the number of timestamps we have to encode), or discarding some of the change history entirely. However, any sort of lossy compression would mean losing the ability to reconstruct the original change history, and thus losing the ability to check the hash chains. Depending on how we use the hash chains (to check integrity, to verify authenticity?), such a trade-off will need to be considered very carefully.

A key insight from the experiments above is that when dealing with large change histories, it is much more efficient to encode the document as a whole than to encode each change separately (in the example above, it makes a factor of 50 difference). This has implications for network protocols that sync Automerge nodes: for example, [Hypermerge](https://github.com/automerge/hypermerge) is currently based on append-only logs of changes, assuming that we're working with one change at a time.

Future protocols will need to detect how far out of sync the two nodes are: if they are almost in sync, it is more efficient to just send the last few missing changes, as each change is typically a few hundred bytes. But if they are far apart (or if one of the nodes is lacking the document entirely) it is more efficient to send the whole-document encoding. Multiple versions of the whole-document encoding can be merged efficiently, without having to turn the document into a log of changes and back again. Essentially, the CRDT can run in two modes: an operation-based mode (where we send changes over the network) and a state-based mode (where we send the whole-document encoding). Network protocols will need to be adapted to take advantage of this choice, by figuring out when to use which mode.

One last thing about the whole-document encoding: while its primary design goal is to be compact, the close second goal is to allow fast loading of the current state from disk. For example, fetching the current value of a particular field (e.g. the title of a document) should not require decoding the whole file, but only reading a bit of metadata and then seeking to the appropriate place in the file. On some platforms, we might even be able to avoid loading the whole file into memory by memory-mapping it and seeking only to those bits that we need.

To test this, I wrote a [hacky experimental decoder](https://github.com/automerge/automerge-perf/blob/master/edit-history/load-text.js) that reads the latest text of the LaTeX paper source example (that is, it extracts the 104,852-character final text string from the 1,119,341-byte encoded file). Here are the timings from running that decoder 10 times on my laptop:

    12.919ms
    6.203ms
    55.147ms
    1.702ms
    1.960ms
    4.270ms
    8.165ms
    5.617ms
    1.977ms
    1.753ms

I don't know why the running time fluctuates so much (GC maybe?), but a median time of ~5 milliseconds is pretty encouraging. Loading a file with hundreds of thousands of changes would probably take minutes with the current Automerge implementation, if it completes at all.

I will write up a detailed specification of the binary format sometime soon. A lot of thought has gone into it, e.g. designing for future extensibility. For now, I want to put this update out there to share what's happening.

By the way, Orion and Alex have been doing an excellent job of tracking this progress in their [Rust port of Automerge](https://github.com/automerge/automerge-rs/tree/performance). This means interop between the JavaScript and Rust backends is not far off!


@types/automerge/index.d.ts
```diff
@@ -203,53 +158,61 @@ declare module 'automerge' {
...
-    diffs?: Diff[]
   }
 
   interface Op {
     action: OpAction
     obj: UUID
-    key?: string
-    value?: any
+    key: string | number
+    insert?: boolean
```
lightsprint09, line 184: For me it is really hard to reason about a optional bool. 
A bool should either true or false. If you really need to express three states, I would prefer a enum with named states.

lightsprint09: This is not the only place in the type specification where optional bool is used. Maybe we could clean this up if possible

scotttrinh: Isn't this a side-effect of writing something in idiomatic JavaScript? Certainly good advice for TypeScript code, but encoding these states as `0, 1, 2` would be pretty opaque to the JavaScript consumer, and using strings would bloat the in-memory cost. In practice, I imagine you'd use a falsy check here, so `undefined` would be equivalent to `false`. I assume this is still just two states (`falsy`, `true`) rather than three (`undefined`, `false`, `true`) in any meaningful sense.

pvh:
Hi Martin, I've had some time now to digest all this and give it some thoughts. Comments inline.

First, on the subject of timestamps, paying for per-millisecond timestamps on all commits is pretty expensive, and it's not clear to me that there is much utility there. I think per-second timestamps would be a quite-reasonable optimization, and I'd further consider making their insertion optional.

Next, as for key movements or other expensive but less valuable data, I've thought quite a bit about "non-dependent actors" as a useful option here. The concept is (in old-automerge lingo) an actor which depends on other actors but is never depended-upon. It could thus track the main document but would not need to be loaded unless it was necessary. Users could provision and synchronize this kind of cursor history during interactive sessions but throw them away afterwards or simply not bother to share them during asynchronous collaboration. It's not clear to me whether this would still be feasible with the new design but it seems in principle like it should be (simply bump the dependency hash as necessary).

As to the future for Automerge Connection I have been aware of this as an upcoming problem for quite some time and have only recently begun giving it serious consideration. The vector clock model is desirable because calculating the necessary synchronization work between two documents is a trivial operation that only needs to examine the clock. On the face of it, calculating a delta between hash-chained commits would be tricky. We'd need to calculate the least-common-ancestor for the peers and then generate a special commit that caught them up. Streaming the changes to allow a client to receive deltas incrementally would be quite expensive in comparison but perhaps it's a desirable feature to send the result in a single exchange.

My mind does wander to the rsync/noms/jump-rope style architecture for synchronization. Is there a mechanical sympathy with a binary-encoding-level synchronization? Could we use some kind of probablistic-blocks method to determine what needs to be exchanged in each direction at the  storage-layer? It's unclear to me right now.

All of this stuff also intersects with privacy concerns. We need to think carefully about how we query between peers and what we expose and to whom. Content hashes are, in general, pretty sensitive.

All that said, I'm very excited that we're getting to the point of working on these problems! I'm eager to dig into all of this in more detail as the binary layers come into greater focus. We'll need to find a way to commission a project to work on this stuff sometime soon, but that's a topic for a different venue.

mlockett42:
Minisketch (https://github.com/sipa/minisketch) might be a useful library for dealing with synchronizing sets of hashes efficiently. This could be more efficient than storing and exchanging a whole lot of vector clock metadata.

HerbCaudill:
> First, on the subject of timestamps, paying for per-millisecond timestamps on all commits is pretty expensive, and it's not clear to me that there is much utility there. I think per-second timestamps would be a quite-reasonable optimization, and I'd further consider making their insertion optional.

+1 on per-second timestamps, and +1 on making timestamps optional altogether.

HerbCaudill:
> The switch to using hash chaining will have a bearing on network protocols that sync up Automerge replicas, such as Automerge.Connection, which currently relies on vector clocks. This is something I would like to discuss with the community. 

For what it's worth, I've got [a lot of code](https://github.com/DevResults/cevitxe/search?q=clock) built on top of the vector clock model; so I'd hate to see it thrown out entirely. I'd understood from [this conversation](https://github.com/automerge/automerge/issues/200#issuecomment-597667817) that the plan was to keep vector clocks but augment them with chain hashing, is that no longer the case? 

> By the way, I have split Automerge.Connection into a separate repository, because I think we might want to evolve it and assign it version numbers independently from the Automerge core.

That definitely strikes me as the right move - it's always seemed a bit out of place in the main codebase.

HerbCaudill:
> A key insight from the experiments above is that when dealing with large change histories, it is much more efficient to encode the document as a whole than to encode each change separately (in the example above, it makes a factor of 50 difference). This has implications for network protocols that sync Automerge nodes: for example, Hypermerge is currently based on append-only logs of changes, assuming that we're working with one change at a time.

Just to be clear, we'll always have access to the individual changes, right? [Cevitxe](http://github.com/DevResults/cevitxe) also stores an append-only log of changes. Maybe there need to be different strategies available for the multiuser-text-editing use case, where you have more individual changes with tiny payloads, vs. the replicated-database use case, where you have fewer changes with more substantial payloads.

@types/automerge/index.d.ts
ept: It's not tristate, it's just a boolean that defaults to false when absent. I don't know how to declare that in TypeScript. In 02271f082803e57ccc2c2d7890fc5ec49ab08f3f I've changed operations such that they always have an `insert` property that is explicitly true or false.

ept:
Okay, I've changed it to use second-resolution timestamps. Here is the updated file size:

| | Binary format | Binary + gzip |
|--|--|--|
| Paper (whole document) | 695,298 bytes | 302,067 bytes |

I am really pleased with this: the gzipped document uses less than one byte per change, while still capturing the full keystroke-by-keystroke change history!

> Just to be clear, we'll always have access to the individual changes, right?

Yes, the full log of individual changes will always be available (e.g. through `Automerge.getChanges()` and `Automerge.getAllChanges()`). The compressed document format is only returned by `Automerge.save()`.

> I've got a lot of code built on top of the vector clock model; so I'd hate to see it thrown out entirely. I'd understood from this conversation that the plan was to keep vector clocks but augment them with chain hashing

Agree, I think we should keep vector clocks but augment them with hash chains. Vector clocks are still useful because they allow two nodes to sync up in one round-trip, while resolving hash chains sometimes requires several round-trips.

ichorid:
May I suggest using `lz4` compression instead of `gzip`? Of course, `lz4`'s compression ratio is considerably worse than that of `gzip`, but the speed of compression is about 6x times higher. Also `lz4` is able to compressed data *incrementally* which is very nice to e.g. [fit as much data as possible into network packet](https://github.com/Tribler/tribler/blob/7c955cb2f6ec6463321b65ecae0436041a464aca/src/tribler-core/tribler_core/modules/metadata_store/orm_bindings/channel_metadata.py#L70) without resorting to backtracking.

ept:
> May I suggest using lz4 compression instead of gzip?

This compression step happens outside of Automerge, so you can combine Automerge with the compression algorithm that best suits your requirements.

dmonad:
I was curious to test the performance improvements in [crdt-benchmarks](https://github.com/dmonad/crdt-benchmarks). This is quite a leap! 

You can compare the results here: https://github.com/dmonad/crdt-benchmarks/pull/4/commits/73f3f650359d12ce22c754aabf36551b438f2658

It is very impressive how well your new algorithm compresses meta information! The only thing that I'm worried about is that parsing of the compressed document `(parseTime)` now takes longer than before. There seems to be quite some overhead in encoding the document in this way. In my experience, the true problem of CRDTs is that they are too slow to parse long edit-histories. Network traffic and storage are not really a problem.

I see that you implemented a similar module for encoding/decoding as I did for [Yjs](https://github.com/yjs/yjs). If you are interested you could make use of [lib0/encoding](https://github.com/dmonad/lib0/blob/master/encoding.js#L1) and [lib0/decoding](https://github.com/dmonad/lib0/blob/master/encoding.js#L1). I already put a crazy effort into optimizing read/write performance.

ankit-m:
@ept I was wondering if you are targeting a release date for v1.

ept:
Hi @ankit-m, no definite schedule, but I hope to have a preview release within the next month or two. A final stable release may take a bit longer (since this branch has a lot of new code, there will be bugs that need to be ironed out) but I'm hoping well before the end of the year.

doctorpangloss:
> two thirds of the file is just timestamps! 

Is it possible to turn off the timestamps?

ept:
> Is it possible to turn off the timestamps?

As you can see from the subsequent comments (which GitHub is now hiding by default), I changed the granularity so that timestamps are now a much smaller fraction of the file. If it's still a problem we can consider allowing it to be turned off, but I'm keen not to add more configuration options to Automerge than really necessary.

mkhanal:
👍 @ept. I think this change is definitely going to shake things up.

With regards to further reduction in size and maintaining history, I'd like to also highlight a use-case that we see and if it is generic enough, maybe it can be a candidate functionality. This is possible because of the introduction of the timestamps on the changes. Also, it takes away the need to make Automerge handle everything as a generic case (high development and maintenance cost) and passes on the responsibility of identifying safety of some of the compressions to the consumers.

### Use case
Timestamps have opened paths for change compression/combination (merging changes) based on the domain (business use case). If the domain does not need history to be preserved forever, a compress call can be created with a timestamp which should allow changes older than timestamp to be compressed/combined into a single change. E.g.

```javascript
Automerge.compress(doc, timestamp1)
// or
Automerge.garbageCollect(doc, timestamp1)
```

### An example
Our application currently runs collaboratively and our document grows to a few MBs in size during a single session. Subsequent sessions increase it further by a similar size. This will continue to happen even after the introduction of the binary data format as our change set will keep growing with every session.

However, apps like these can chose not to allow collaborators to drift for longer than a certain period of time without syncing (few minutes to a few days). So any changes after `T` time period can be dropped. For this reason, we know we don't need to keep history that are older than `T` time period. Compression like this can even happen on the background periodically as a mark and sweep operation (like running a garbage collection) and should be acceptable to be slower than normal.

Does this sound like a viable solution for further compression? I understand that we want to stay away from configurations in Automerge for this, hence an explicit API. 

Happy to start a separate thread if this does not belong here.

SahidMiller:
@mkhanal I've opened an issue #277 that seems close to what you're requesting but in regards to the master branch rather than the performance branch.

Since opening the issue, I was able to modify Automerge internals to properly rehydrate large opsets from a serialized representation of the *result* of the change (essentially recursively calling .toJSON and adding _objectId and _type and other properties needed) and then calling Automerge.from + Automerge.getChanges to rehyrdrate the opset (a performance hit for doing it this way)

It's not exactly performant but it does save space over the wire and on disk... but it only works for specific cases where most if not all the objects in the change were created in that change. It's not generic enough to catch more complex changes, for example, modifying objects created outside that change. Although, I don't see why it's not possible, just a bit difficult and probably unnecessary for my application. If you're interested, I can share some code snippets with you.

Valodim:
For the record, history scrubbing has been discussed a bit in #259 before. I would personally also love to see a history scrubbing operation, but I suspect it's a bit out of scope for this PR. :)

weslien:
Just curious, I haven't read enough about Automerge and the new binary protocol to judge for myself, but wouldn't Apache Arrow be something you should look into? It's not really the same thing, but I think it might be something that would complement your efforts, and it also brings its own set of benefits. The combination, at least in my mind, is quite tantalizing.

ept:
@weslien I am aware of Apache Arrow. It's a good project, but it's designed for data analytics, which has somewhat different characteristics from the collaborative editing data managed by Automerge. The columnar encoding in Automerge has some features specific to its needs, which would not be met by a generic format. Hence I think it does make sense for Automerge to define and implement its own format.

beorn:
Really looking forward to the new binary format!

I don’t want to hijack this thread, but I have some questions regarding the storage format which would probably be impacted by any format change:

How easy/feasible will it be to make the storage layer a pluggable, so that you could have a more granular control over how things are persisted, e.g., persist changes or even diffs in a database?  (Or as in the case of Hypermerge, each log in a different Hypercore - so I guess it’s somewhat possible now.)

Or would this not make sense anyways since Automerge would need to load the entire dataset into memory anyways at startup and/or sync/diff/merge time to do the CRDT operations?

The use cases I’m thinking about are scaling to a syncing server that is used to exchange changes with a lot of users (and so may not need to keep the entire doc loaded, just the recent changes), and a database-like usage of Automerge where there are a lot of objects in it (so you may not fit the entire database in memory).

ept:
Hi @beorn, this is a good question. My intention is to keep the storage layer separate from Automerge. That is, Automerge's API consumes and returns byte arrays, but whether those byte arrays are stored consecutively in a flat file, or as rows in a database, or as records in a hypercore, etc. is up to you to decide. I expect different apps will want to make different decisions in this regard.

There are two encoding formats: the document format, where the whole document (and its editing history) is encoded in a single byte array, and the change format, where each change is encoded as a separate byte array. The two have the same information content: the log of all individual changes can be turned into a document, and a document can be turned back into its constituent changes. The document format is more compact, making it more suitable for bringing a new client up-to-speed and for storing the state of a document on disk. On the other hand, the document format is rewritten on every change, so if you want to send some small incremental updates with low latency, the change format is better. You can combine the two: for example, you could write out a snapshot in the document format once an hour or every 1,000 changes, and use the change format for incremental updates in between snapshots. That way you get the best of both formats.

For a storage layer, the bottom line is that in order to get good performance, you might want to distinguish between these two formats. However, the storage layer does not need to know anything else about the encoding, other than the distinction between a whole document and a log of changes.

beorn:
Thanks Martin, that sounds like the perfect solution to the syncing / saving use cases!

And this maybe belongs somewhere else - but how about the CRDT apparatus of Automerge itself and its working set memory requirements:

Does it always require everything loaded into memory, or can it lazily load and use changes?  (The change-exchange-hub/server use case.)

Is there be any way to lazily load parts of the object tree?  I guess a la Yjs sub-documents or even finer granularity at the object/prop level.  (The Automerge-as-a-big-database use case.)  I guess it’d be okay to even support this for Automerge.Table data types, if that makes it easier.

ept:
At the moment Automerge assumes that a whole document will be loaded into memory at a time. If you have large amounts of data, I suggest breaking it up into several smaller documents. The binary format for a document is designed to potentially support partial loading in the future (data relating to the same object is grouped to be close together, so we get some locality benefits), and I could imagine that in some systems, you may e.g. mmap a file containing a large document, and then read only the parts of it that are of interest by relying on the OS page cache. Obviously this will only be possible on platform where the necessary syscalls are accessible, i.e. not in a web browser. But I want to get this performance branch finished and released before we go too far into partial loading.

beorn:
Thanks Martin!  Great to hear that partial loading is somewhere on the roadmap!

atifsyedali:
> @weslien I am aware of Apache Arrow. It's a good project, but it's designed for data analytics, which has somewhat different characteristics from the collaborative editing data managed by Automerge. The columnar encoding in Automerge has some features specific to its needs, which would not be met by a generic format. Hence I think it does make sense for Automerge to define and implement its own format.

@ept I am wondering if you have looked into [Apache Parquet](https://parquet.apache.org/documentation/latest/). While Apache Arrow (and the Arrow IPC file format, as used by most examples on the Arrow docs) is a bit apples-oranges comparison to your encoding work as Arrow is more about efficient in-memory representation for analytics use cases, Parquet focuses solely on the efficiency of storage, and uses many techniques including RLE. 

There are also several implementations of Parquet, including a [JS](https://github.com/ironSource/parquetjs) and [Rust](https://github.com/sunchao/parquet-rs) one, which would significantly improve time-to-port to other languages.

You can also write/read row groups, which would give you partial saving and loading for large files. Or the ability to order your rows chronologically so you can stop reading at a defined point (given there is some rollup snapshot in between the row groups).

ept:
@atifsyedali Yes, I have looked into Parquet, and similar comments hold for it as for Arrow: although some of the basics like RLE are similar, Automerge has specific requirements that these projects don't meet. For example, the details of encoding a text CRDT efficiently go beyond what they provide, because analytical databases simply don't have this type of data. Another issue is that formats designed for analytics are designed to be mostly read, and writes/updates happen in bulk. On the other hand, Automerge needs to process lots of small, frequent, incremental updates (e.g. keystrokes in a text document), and apply these to the document state efficiently. This requires a lot of control over the data structures used.

An argument in favour of using a general-purpose data format would be that there is existing tooling for it, to inspect it and write custom tools that use the data. However, even if Automerge used Parquet, I think the output would look so low-level and confusing that it would be difficult to work with. Therefore, I think the right approach is for us to develop good tooling around a custom data format.

danny-andrews:
Super exciting stuff! I haven't looked over the PR closely, so sorry if this is redundant, but one thing that would be nice to have is a straightforward API for converting the binary format into something inspectable. I've been playing around with yjs, and one thing that's been frustrating to me is figuring out how to convert updates into something human readable, both to facilitate learning how it works, and for doing things like creating my own persistence layer. I haven't found any public API in yjs to achieve this.

Cheers.

ept:
> one thing that would be nice to have is a straightforward API for converting the binary format into something inspectable

I agree. There are already functions to convert between the binary format and an equivalent JSON representation (these are extensively used in the tests, for example), and we should make this part of the public API.

backend/index.js
```diff
@@ -1,283 +1,374 @@
...
-  for (let change of fromJS(changes)) {
-    change = change.remove('requestType')
-    const [newOpSet, diff] = OpSet.addChange(opSet, change, undoable)
-    for (let d of diff) diffs.push(d)
-    opSet = newOpSet
+function apply(state, changes, request, isUndoable, isIncremental) {
+  let diffs = isIncremental ? {} : null
+  let opSet = state.get('opSet')
+  for (let change of changes) {
+    for (let chunk of splitContainers(change)) {
```
nornagon, line 164: `chunk` is never used inside the loop... does this only work because the changes only contain one chunk by accident?

ept: You're right, it should be using `chunk` instead of `change` inside the loop. I've pushed e3d0170b1249ed401cf38713776925aad981e6e9 to fix this. Thanks for pointing it out!

@types/automerge/index.d.ts
```diff
@@ -204,76 +177,61 @@ declare module 'automerge' {
...
   }
 
-  type RequestType =
-    | 'change' //..
-    | 'redo'
-    | 'undo'
+  interface Edit {
+    action: 'insert' | 'remove'
+    index: number
+    elemId: OpId
```
lightsprint09, line 222: Is elemId really non optional or is there missing a `?`

At the following there is an edit created which does not includes an elmId.
https://github.com/automerge/automerge/blob/81079ff75d2234b47cb912bad728158f2e71c527/frontend/context.js#L424

jeffa5: I believe it is optional for removes, the rust version may help https://github.com/automerge/automerge-rs/blob/c9ba656c0c7947c948b83a6d9432eb93e26c825b/automerge-protocol/src/lib.rs#L389-L398

ept:
All the compatibility-breaking changes that I've planned to make are now completed and merged into the performance branch, and so I'm declaring this branch done and merging it into main. No doubt there will be bugs, but we can address them on separate PRs. Hooray, it's shipping! 🚢

PR #261, opened 2020-06-13
support calling indexOf with an object

philschatz:
The parameter to `[].indexOf(...)` may be a Proxy object. In that case, the correct index should be returned. Currently it always returns `-1`.


ept:
Thank you @philschatz for spotting this bug, and for the nice fix!

PR #265, opened 2020-06-26
Performance: add typescript src

NikitaIT:
The proposed changes are an intermediate step for switching to typescript.
For a more trivial merge, the contents of the files were not changed (only necessary for compilation).

- Changes do not affect the bundle.
- Changes do not affect index.d.ts.
- Changes increase karma test time.

**DX:** 

- noImplicitAny: false - for change extension without code changes
- allowJs - for import not typed backend without errors

**Chore:** 

- add .idea for webstorm

Perhaps there are ways to make running tests better, but I have too little karma for this.

Oh, Safari not works on Windows.

Wait for typings.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@
 /node_modules
 .nyc_output
 .vscode
+/.idea/
--- a/frontend/apply_patch.ts
+++ b/frontend/apply_patch.ts
@@ -304,6 +304,6 @@ function cloneRootObject(root) {
   return cloneMapObject(root, ROOT_ID)
 }
 
-module.exports = {
+export default {
   interpretPatch, cloneRootObject
 }
--- a/frontend/apply_patch.ts
+++ b/frontend/apply_patch.ts
@@ -304,6 +304,6 @@ function cloneRootObject(root) {
   return cloneMapObject(root, ROOT_ID)
 }
 
-export default {
+export = {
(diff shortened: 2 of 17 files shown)


NikitaIT:
# Ts

Typescript is hard to use with require. Need change `export =` to `export` with types. 
Maybe it's worth building it as separate bundles.

Are there any restrictions on the typescript language features that can be used?

# Questions about next steps

## Common

Its just utils, no cohesion, only low coupling. Import as single module should not exists.

### isObject(obj)

Maybe it's guard for copyObject assumes.

### copyObject(obj)

Assumes obj can't have defineProperty for fast copy.

But vue.js make all as defineProperty. Maybe a problem?

### parseOpId(OpSerializedId | null | undefined)

Why null | undefined allowed?

### equalBytes(Uint8Array, Uint8Array)

Only Uint8Arrays? What does it mean? If this is a domain restriction, why is it checked in this utility method.

RealType of equalBytes is EqualBytesArray<T>:

```typescript
type EqualBytesArray<T> = {
  byteLength: number;
  [index: number]: T; // where T in Eq<T>
};
```

### ROOT_ID

It shuld be only guid structure? RealType of ROOT_ID should be defined immediately for all same types, I think.

### uuid

Why is not a class?

### init

Why init has equal error assertion as Frontend.init. It might be worth putting it in a single method.

PR #279, opened 2020-10-23
Move: first move semantic for text

MeneDev:
Experimenting with move semantics

Current status:

* simple moves of single elements seem to work now
* new action 'mov' to encode moves in a single action
  * uses `key` and `child` to encode operation
  * moves from child to key

Untested:

* multiple moves 
* concurrent moves
* anything more complex than moving a single character once

Still figuring out:

* moves are encoded as insert/remove on edits level, not sure if that works in the general case

Planned:

* somehow define a range of elements to create an atomic move of a whole range

### DIFF
--- a/backend/op_set.js
+++ b/backend/op_set.js
@@ -70,6 +70,42 @@ function applyInsert(opSet, op) {
     .setIn(['byObject', objectId, '_insertion', opId], op)
 }
 
+function applyMove(opSet, op, patch) {
+  const objectId = op.get('obj'), opId = op.get('opId'), key = op.get('key'), child = op.get('child')
+  if (!opSet.get('byObject').has(objectId)) throw new Error(`Modification of unknown object ${objectId}`)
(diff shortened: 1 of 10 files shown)


backend/op_set.js
```diff
@@ -587,6 +631,18 @@ function insertionsAfter(opSet, objectId, parentId, childId) {
     .map(op => op.get('opId'))
 }
 
+function movesAfter(opSet, objectId, parentId, childId) {
```
MeneDev, line 634: this is unused

backend/op_set.js
```diff
@@ -604,12 +660,35 @@ function getNext(opSet, objectId, key) {
...
+    const beforeParent = getPrevious(opSet, objectId, parentId)
+    // potentially the element can be moved around several times (or even just once)
+    // and end up in the same spot again
+    const moveOp = moves.last()
+    if (moveOp.key === beforeParent) {
+      break
+    }
+
+    parentId = beforeParent
+  }
```
MeneDev, line 682: unused

PR #282, opened 2020-11-18
Add global "Automerge"

infiniteluke:
This will allow usage with unpkg cdn in a browser environment that has no build system, for example.


infiniteluke:
Closing as this is covered by https://github.com/automerge/automerge/blob/main/webpack.config.js#L8 already and my initial test was bad

PR #288, opened 2020-11-26
Allow hooking into patch application

ept:
Here is a simple but potentially very important feature: allowing applications to hook into Automerge's process for applying patches to a document.

The use case is when you have some data outside of Automerge that you want to keep consistent with the state of an Automerge document. This could be a user interface (especially a complicated widget like a text editor), or a cache, a search index, or suchlike. Such an application needs a mechanism to get notified when an Automerge document changes, so that it can apply the appropriate update to the external state.

At the moment, Automerge only really supports the React-style approach where you pass the entire document to the render function, and the renderer works out which bits of the document have actually changed, and updates those, while leaving the unchanged parts untouched. But in other circumstances, it would be really useful to have a description of the change that occurred, not just the before-and-after document. Moreover, this process should be the same, regardless of whether a change is generated by the local user or was received over the network from a remote user.

Automerge's existing concept of a patch is probably the best form we have for describing what change occurred. It's easier for applications to consume than the raw change format because it includes more context, and e.g. for list updates it includes the index of the list position being changed, not just the list element ID. This patch format is currently used for updates sent from the backend to the frontend, and used by the frontend to update the document.

This PR adds the ability for applications to get a callback for every patch that is applied to a document. When making a local change, you can add a callback like this:

```js
newState = Automerge.change(oldState, {patchCallback: (patch) => {...}}, doc => {
  // mutate doc as usual
})
```

And when applying a remote change, you can add a callback like this:

```js
newState = Automerge.applyChanges(oldState, changes, {patchCallback: (patch) => {...}})
```

The patch format is the same in both cases. The patch format used by the currently released version of Automerge is documented in [`INTERNALS.md`](https://github.com/automerge/automerge/blob/0678f39eccb9043c5b0ec2189ead49115d88449c/INTERNALS.md#patch-format) (in the "Frontend-backend protocol" section). This format has changed considerably on the `performance` branch, and the new format is documented in [`BINARY_FORMAT.md` on that branch](https://github.com/automerge/automerge/blob/4d605e526f5da74a226e3fdf641b6ff5d0705213/BINARY_FORMAT.md#patch-format). The new format is intended to be easier to consume.

Note that this API is not intended to allow applications to modify the patches before they get applied. The patch callback should treat the patch as read-only.

An alternative API would be to pass in a callback function once when initialising the document, and automatically calling it on all subsequent updates. But I thought that had a bit of a "spooky action at a distance" feel, while the callback on every `change`/`applyChanges` call was more explicit. Though it could also be annoying for app developers to have to add the option to every single `Automerge.change` call, which might be scattered around an application. Any thoughts on this?

This feature is especially for @cklokmose, but I welcome feedback from others as well. @HerbCaudill what do you think?

### DIFF
--- a/frontend/index.js
+++ b/frontend/index.js
@@ -109,6 +109,7 @@ function makeChange(doc, requestType, context, options) {
 
   if (doc[OPTIONS].backend) {
     const [backendState, patch] = doc[OPTIONS].backend.applyLocalChange(state.backendState, request)
+    if (options && options.patchCallback) options.patchCallback(patch)
     state.backendState = backendState
     state.requests = []
     return [applyPatchToDoc(doc, patch, state, true), request]
--- a/src/automerge.js
+++ b/src/automerge.js
@@ -95,9 +95,10 @@ function getAllChanges(doc) {
   return getChanges(init(), doc)
 }
 
-function applyChanges(doc, changes) {
+function applyChanges(doc, changes, options = {}) {
   const oldState = Frontend.getBackendState(doc)
   const [newState, patch] = Backend.applyChanges(oldState, changes)
+  if (options.patchCallback) options.patchCallback(Object.assign({}, patch))
   patch.state = newState
   return Frontend.applyPatch(doc, patch)
 }
--- a/src/automerge.js
+++ b/src/automerge.js
@@ -11,6 +11,7 @@ function docFromChanges(options, changes) {
   const doc = init(options)
   const [state, _] = Backend.applyChanges(Backend.init(), changes)
   const patch = Backend.getPatch(state)
+  if (options && options.patchCallback) options.patchCallback(Object.assign({}, patch))
   patch.state = state
   return Frontend.applyPatch(doc, patch)
 }
--- a/test/test.js
+++ b/test/test.js
@@ -1378,6 +1378,27 @@ describe('Automerge', () => {
       doc = Automerge.load(Automerge.save(doc))
       assert.deepEqual(doc.foo, [1])
     })
+
+    it('should call patchCallback if supplied', () => {
+      let patches = []
+      let s1 = Automerge.change(Automerge.init(), doc => doc.birds = ['Goldfinch'])
+      let s2 = Automerge.change(s1, doc => doc.birds.push('Chaffinch'))
+      Automerge.load(Automerge.save(s2), {patchCallback: p => patches.push(p)})
+      assert.deepStrictEqual(patches, [{
+        canUndo: false, canRedo: false,
+        clock: {[Automerge.getActorId(s1)]: 2}, deps: {[Automerge.getActorId(s1)]: 2},
+        diffs: [
+          {action: 'create', obj: Automerge.getObjectId(s1.birds), type: 'list'},
+          {action: 'insert', obj: Automerge.getObjectId(s1.birds), type: 'list',
+            elemId: `${Automerge.getActorId(s1)}:1`, index: 0, value: 'Goldfinch'},
+          {action: 'insert', obj: Automerge.getObjectId(s1.birds), type: 'list',
+            elemId: `${Automerge.getActorId(s1)}:2`, index: 1, value: 'Chaffinch'},
+          {action: 'maxElem', obj: Automerge.getObjectId(s1.birds), type: 'list', value: 2},
+          {action: 'set', obj: ROOT_ID, type: 'map', key: 'birds',
+            value: Automerge.getObjectId(s1.birds), link: true}
+        ]
+      }])
+    })
   })
 
   describe('history API', () => {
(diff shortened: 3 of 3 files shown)


frontend/index.js
```diff
@@ -109,6 +109,7 @@ function makeChange(doc, requestType, context, options) {
 
   if (doc[OPTIONS].backend) {
     const [backendState, patch] = doc[OPTIONS].backend.applyLocalChange(state.backendState, request)
+    if (options && options.patchCallback) options.patchCallback(patch)
```
lightsprint09, line 112: I would prefer a operate line + indentation, which improves readability, safety and debuggablility.

src/automerge.js
```diff
@@ -11,6 +11,7 @@ function docFromChanges(options, changes) {
   const doc = init(options)
   const [state, _] = Backend.applyChanges(Backend.init(), changes)
   const patch = Backend.getPatch(state)
+  if (options && options.patchCallback) options.patchCallback(Object.assign({}, patch))
```
lightsprint09, line 14: Same as above

src/automerge.js
```diff
@@ -95,9 +96,10 @@ function getAllChanges(doc) {
   return getChanges(init(), doc)
 }
 
-function applyChanges(doc, changes) {
+function applyChanges(doc, changes, options = {}) {
```
lightsprint09, line 99: why is there an options map needed. Could we just pass the callback directly?

src/automerge.js
```diff
@@ -95,9 +96,10 @@ function getAllChanges(doc) {
...
   const oldState = Frontend.getBackendState(doc)
   const [newState, patch] = Backend.applyChanges(oldState, changes)
+  if (options.patchCallback) options.patchCallback(Object.assign({}, patch))
```
lightsprint09, line 102: I would prefer a operate line + indentation, which improves readability, safety and debuggablility.

lightsprint09 (approved):
Overall I really like this feature.

src/automerge.js
HerbCaudill: I prefer the interface @ept suggests here - the options map makes it easier to add the change API further in the future. It also makes the code calling this function more readable.

HerbCaudill:
Overall this seems like a step in the right direction. It always seemed odd to me that you could watch for changes on a `DocSet` but not on an individual document. 

> An alternative API would be to pass in a callback function once when initialising the document, and automatically calling it on all subsequent updates. But I thought that had a bit of a "spooky action at a distance" feel, while the callback on every change/applyChanges call was more explicit. Though it could also be annoying for app developers to have to add the option to every single Automerge.change call, which might be scattered around an application. Any thoughts on this?

I'd imagine most developers would find the idea of a change listener to be totally natural, and not spooky at all. 

Is there a reason why we can't have both? 

```js
const patchCallback = patch => alert(patch)
const changeFn = doc => { doc.foo = 42 }

// this works
const state = Automerge.init({}, { patchCallback }) // patchCallback will fire on every change
const newState = Automerge.change(state, changeFn)

// this works too
const state = Automerge.init({})
const newState = Automerge.change(state, changeFn, { patchCallback }) // patchCallback only fires on this change
```

HerbCaudill:
On second thought I do see that this is a little spooky since you're getting a new Automerge document every time. 

```js
const state = Automerge.init({}, { patchCallback }) 
const state1 = Automerge.change(state, s => s.foo = 42)
const state2 = Automerge.change(state, s => s.boo = 'pizza') // maybe it is weird for patchCallback to fire here
```

src/automerge.js
lightsprint09: I agree that readability would improve on the call side. But I think that's caused by missing labeled arguments in JavaScript which we should not try to fix.

When just reading `function applyChanges(doc, changes, options = {})` I would expect to pass a document, some changes and a map which only contains key **value** pair, which help configuring the change method. I would not expect to pass references into `options` which provide me with the resulting patches.

On the other hand when reading `function applyChanges(doc, changes, patchCallback)` it is clear to me that I can pass a document, changes and a callback which provides some kind of patches. (It is also future prove to just add another parameter)

This is not really a strong opinion, so I am happy to go with options map as well :)

ept: I'm with @HerbCaudill here. I think this kind of options hash is common in JavaScript, since it gives us something a bit like named parameters in the absence of language support, and it allows us to add further options in the future without breaking existing code. Moreover, lots of existing functions in the Automerge API (`init`, `from`, `load`, `change`, etc.) already take an options hash. I don't see a problem with passing a function as part of an options hash: in JS, functions are values like any other.

cklokmose:
Thanks @ept that is very useful for my use case. It will simplify the code receiving remote changes and applying them both to the Automerge document and the user facing representation of the document.

ept:
This PR is based on the Automerge 0.x release series. As we are now focussing our attention on the 1.x series, I will close this PR. #308 is the equivalent of this PR for 1.x.

PR #301, opened 2021-01-11
frontend: splice's deleteCount can overflow

pierreprinetti:
Before to this patch, calling `splice` on an Automerge list with a
deleteCount that overflows the list results in a RangeError.

However, according to Mozilla's documentation (and tested on Node
v14.15.4), splice's `deleteCount` argument can be equal or greater than
the list. In that case, all the elements from start to the end of the
array should be deleted.

With this patch, deleteCount can overflow the list's length and the
behaviour is like no deleteCount was passed.

[1]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/splice

### DIFF
--- a/frontend/proxies.js
+++ b/frontend/proxies.js
@@ -75,7 +75,7 @@ function listMethods(context, listId) {
     splice(start, deleteCount, ...values) {
       let list = context.getObject(listId)
       start = parseListIndex(start)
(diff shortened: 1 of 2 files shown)


ept:
Thank you @pierreprinetti, this is a nice contribution!

Issue #303, opened 2021-01-15
removeDoc does not clean up connections' _ourClock caused re-inited document out of sync.


ulion:
automerge have DocSet, and Connections, when any change do any doc happens, it call the handlers (docChanged) of every connections, which then spread the clock of one doc to every other connections no matter which doc is for those connections. For now, it still looks good, if outside frame does not want send unrelated doc clocks, it can filter it out like https://github.com/cudr/slate-collaborative/commit/ab3c36ab7dd7f24059c5d0864735c5892df98028#r45985664 did. But as I commented there, the removeDoc of DocSet is unclean, the clock of the removed doc are still in every other connections' _ourClock cache. which cause the re-inited doc stop sync, and the well known "Unexpected error in receiveOperation RangeError: Cannot pass an old state object to a connection" error.

It is much more related to automerge, not the upper caller. the problem is, what's the clean way to fix this? @ept

ept:
Hi @ulion, thanks for this. On the 1.x release series of Automerge the Automerge.Connection and Automerge.DocSet APIs have been removed, and we are replacing them with new APIs (see #290). We plan to make the first 1.0-preview release soon, and for this reason I am not planning to maintain Automerge.Connection any more.

Issue #290, opened 2020-12-02
Automerge 1.0 sync protocol


ept:
Hi everyone, I would like to gauge opinions on a design decision that needs to be made for the `performance` branch and Automerge 1.0. It is particularly relevant for people working on networking layers for Automerge, such as [Hypermerge](https://github.com/automerge/hypermerge) (@pvh) and [Cevitxe](https://github.com/DevResults/cevitxe) (@HerbCaudill).

We have long talked about using Git-style hash chaining to capture dependencies between changes (#27, #200), instead of the current system of referencing dependencies by their actorId and sequence number. This avoids bugs due to applications inadvertently reusing actorIds or sequence numbers (e.g. #234). Hash chaining is now implemented as part of the compressed binary format (#253), and it seems to be working well in terms of document storage.

However, we now need to figure out the impact on Automerge networking layers. When two peers connect to each other, or when a client connects to a server, they need to figure out which changes they need to send each other in order to get in sync. At the moment, this is typically done using vector clocks: that is, a mapping from actorIds to sequence numbers, indicating the highest sequence number we've seen from a given actorId. If you receive a vector clock and you have some changes with higher sequence numbers for a given actorId, you send those changes to the other node. This makes for nice and simple protocols. It also generalises to variants: for example, Hypermerge uses a [Hypercore](https://hypercore-protocol.org/) (an append-only log) per actorId.

I am keen to keep the core Automerge library network-protocol-agnostic, so that it can be paired up with different networking layers depending on the needs of particular applications. But we do need to figure out the core APIs that Automerge needs to provide in order to allow the networking layers to do their work. For example, #284 adds an explicit vector clock API (previously you had to mess around with the internal data structures to get the vector clock). What do we want this API to look like for Automerge 1.0?

1. One option is to keep vector clocks mostly as they are. This is possible because the binary format still has actorIds and sequence numbers — they are just no longer being used to track dependencies between changes.
2. Alternatively, we can provide a different API that networking layers can use to figure out what changes to send, which is based solely on hashes. This can be done: @heidihoward and I just published a paper containing algorithms to do just that ([paper](https://arxiv.org/abs/2012.00472), [blog post](https://martin.kleppmann.com/2020/12/02/bloom-filter-hash-graph-sync.html)).

Advantages of vector clocks:

[...]

PR #304, opened 2021-01-15
Only spread doc clock to a connection if the remote care the doc.

ulion:
Trying to fix https://github.com/automerge/automerge/issues/303
connection should not cache other doc's clock if the doc is unrelated (not in _theirClock). else removed then re-loaded doc will have conflict new clock with the cached one. which will cause the re-loaded doc not sync, and raise "old state" error.

### DIFF
--- a/src/connection.js
+++ b/src/connection.js
@@ -67,9 +67,8 @@ class Connection {
         this.sendMsg(docId, clock, changes)
         return
       }
+      if (!clock.equals(this._ourClock.get(docId, Map()))) this.sendMsg(docId, clock)
     }
-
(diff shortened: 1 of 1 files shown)


ulion:
Well, this is only a temp fix. Indeed, no matter whether theirClock interested or not the docId, when a doc is removed from DocSet, it should be also removed from all connection's _ourClock. But I don't have idea how to do it easily.

ept:
Hi @ulion, thanks for this. On the 1.x release series of Automerge the Automerge.Connection API has been removed, and we are replacing it with new APIs (see #290). We plan to make the first 1.0-preview release soon, and for this reason I am not planning to maintain Automerge.Connection any more.

Issue #198, opened 2019-07-31
Exporse oppsite of .getElemId(index)


Gozala:
As [per suggestion](https://github.com/automerge/automerge/issues/193#issuecomment-516980712) in #193 

> The Text class has a `.getElemId(index)` method that looks up the element ID for a particular index; the translation in the opposite direction doesn't seem to be exposed in the API at the moment, but that would be easy to add.

I attempted to add `.getElemIndex(id)` method to `Text.prototype` but I'm afraid I failed to figure out how to do it. `contex.js` does not seem to have anything relevant as far as I can tell.

I would love to help address this but I'd need some mentoring to get it done.

ept:
Oh crap. I had thought it would be easy, but I misremembered some of the details.

Well, there is an easy method, which is to do a linear scan over the `this.elems` array in Text, and to find the index of the element whose `.elemId` property matches. But that's O(n) slow.

What I had in mind was to use the `indexOf` method of the `SkipList` class in the backend, which does the same in O(log n) time. But since the Text object lives in the frontend, and the frontend is not supposed to directly access data structures in the backend (they are supposed to communicate only by message-passing), that method can't be used directly.

It's annoying that the frontend/backend split is making this difficult. Perhaps we need to put a better data structure like the skiplist in the frontend, but I had been trying to keep the frontend as lightweight as possible, otherwise what's the point of splitting the two?

Gozala:
> Oh crap. I had thought it would be easy, but I misremembered some of the details.
>
> Well, there is an easy method, which is to do a linear scan over the this.elems array in Text, and to find the index of the element whose .elemId property matches. But that's O(n) slow.
 
That’s actually not going to work either, because deleted characters are no longer in there

ept:
> deleted characters are no longer in there

Oh yes, you're right. I think the backend skiplist retains deleted element IDs and calculates indexes correctly (deleted elements can be looked up by element ID, but don't count towards indexes), but I'm still not sure how to give the frontend access to that information.

Gozala:
I'm pretty lost in the code base, can't seem to figure out where do frontend / backend even communicate. 

@ept would you mind providing a bit more insights on how I can go about this ? At this point I'm open to some dirty hacks so I can make more progress on my editor prototype & learn more about how things work internally in the process.

ept:
[...]

PR #306, opened 2021-01-24
WIP: initial implementation of cursors

geoffreylitt:
An attempt at a simple initial implementation of cursors, as discussed in #239. 

- For now I just used a plain object to store the Cursor, and added a function `Automerge.Frontend.findCursorIndex` which operates on that object, because that was easiest to implement. Probably would provide a nicer API to switch it to a custom datatype?
- I first tried a pure frontend implementation (preserved in commit a80bfe1) but I believe this is limited in how it can handle deleted characters, as mentioned in #198. So I then switched to using the backend state to find the closest index in case of deleted characters, based on [your suggestion here](https://github.com/automerge/automerge/issues/198#issuecomment-519244431). This at least avoids any changes to the frontend-backend protocol, but I wonder if there are still other issues caused by involving the backend. In particular, I'm not sure how this change affects frontend perf yet.
- It seems like the best handling of deleted characters might depend on context of how the cursor is being used. In this PR I provided an option for how to handle the case where the character has been deleted (either return index -1, or try to find the closest one), but not yet sure if this is the best API.
- I wonder if it would make sense to provide this functionality for both lists and text?

### DIFF
--- a/frontend/index.js
+++ b/frontend/index.js
@@ -3,6 +3,7 @@ const { ROOT_ID, isObject, copyObject } = require('../src/common')
 const uuid = require('../src/uuid')
 const { applyDiffs, updateParentObjects, cloneRootObject } = require('./apply_patch')
 const { rootObjectProxy } = require('./proxies')
+const { findClosestListIndex } = require('../backend/op_set')
 const { Context } = require('./context')
 const { Text } = require('./text')
 const { Table } = require('./table')
@@ -492,10 +493,34 @@ function getElementIds(list) {
   return list[ELEM_IDS]
 }
 
+/**
+ * Finds the latest integer index of a cursor object.
+ * If the character was deleted, returns -1.
+ * 
+ * todos:
+ * - consider returning the closest character if deleted
+ * - consider optimizing the linear search
+ */
+function findCursorIndex (cursor, doc, findClosest) {
+  if(cursor.textId === undefined || cursor.elemId === undefined) {
+    throw new TypeError('Invalid cursor object')
+  }
+
+  const backend = getBackendState(doc)
+  const elemIds = backend.getIn(['opSet', 'byObject', cursor.textId, '_elemIds'])
+  let index = elemIds.indexOf(cursor.elemId)
+
+  if (index === -1 && findClosest) {
+    index = findClosestListIndex(backend.getIn(['opSet']), cursor.textId, cursor.elemId)
+  }
+
+  return index
+}
+
 module.exports = {
   init, from, change, emptyChange, applyPatch,
   canUndo, undo, canRedo, redo,
   getObjectId, getObjectById, getActorId, setActorId, getConflicts,
-  getBackendState, getElementIds,
-  Text, Table, Counter
+  getBackendState, getElementIds, findCursorIndex,
+  Text, Table, Counter,
 }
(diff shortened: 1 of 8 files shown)


frontend/index.js
```diff
@@ -492,10 +493,34 @@ function getElementIds(list) {
...
 
+/**
+ * Finds the latest integer index of a cursor object.
+ * If the character was deleted, returns -1.
+ * 
+ * todos:
+ * - consider returning the closest character if deleted
+ * - consider optimizing the linear search
+ */
+function findCursorIndex (cursor, doc, findClosest) {
```
geoffreylitt, line 504: I think it's odd that this function requires `doc` as a separate argument, but I couldn't find any way to navigate from an object within a doc to the root doc object.

ept:
Thanks @geoffreylitt, this is a good start! I pushed a few commits to change some details. In particular, I want to avoid the frontend calling directly into the backend, since that will cause them to become much more closely coupled than they should be.

I moved the the `findCursorIndex` function to the top-level Automerge module, which is where the integration of frontend and backend takes place. I renamed it to `getCursorIndex` and swapped its argument order for consistency with the rest of the API. I also made the API exported by the backend explicit.

We will need to make the equivalent change in the Rust implementation to maintain compatibility between the two implementations. This should not be hard.

I will call out that this approach to cursors is incompatible with running frontend and backend on separate threads, since it relies on synchronously calling into the backend. I think that's probably an okay trade-off to make since most Automerge users probably aren't using multiple threads, and having this direct call into the backend is a lot simpler than updating the frontend-backend protocol. But weakening our support for multiple threads is something we should discuss, and make an explicit decision about.

alexjg:
This is great work, I'm glad someone is taking on the cursor challenge :slightly_smiling_face: 

I'm also concerned about the explicit coupling to the backend. If we rule out the use of different threads for the frontend/backend protocol (and we would be ruling it out in Rust, the Backend and Frontend components have not been designed to be threadsafe and would require a significant rewrite to enable it) then we lose a lot of advantages the design gives us. I'm currently working on performance for the Rust codebase and aiming to make any change that happens on the UI thread fit within a 16ms budget in order to make 60FPS UI possible, removing multithreading makes this extremely difficult.

Is it possible to achieve this same interface by adding an index to the frontend for the element ID <-> list index rather than asking the backend for that information?

frontend/text.js
```diff
@@ -27,6 +27,18 @@ class Text {
     return this.elems[index].elemId
   }
 
+  /**
+   * Returns a cursor that points to a specific point in the text.
+   * For now, represented by a plain object.
+   */
+  getCursorAt (index) {
+    return {
+      // todo: are there any points in the lifecycle where the Text object doesn't have an ID?
```
geoffreylitt, line 36: @ept I see some [other places](https://github.com/automerge/automerge/blob/3e72b69bafb44d6625496b9d956fdad181530006/frontend/text.js#L116) that handle the case where the object ID is not set.

Would it be sensible to add a similar check here and forbid getting a cursor if there's not an object ID? I don't yet understand when/how that case occurs.

geoffreylitt:
@ept thanks, all those updates make sense.

re: frontend/backend split -- as an alternative to @alexjg 's suggestion of maintaining more state in the frontend, what if `getCursorIndex` was a function exposed by the backend, and the user could choose to either: 1) synchronously call the backend as in the current PR, 2) wire up their own messaging to make an async call.

I guess one problem might be that until now the protocol has entirely consisted of change requests going one way and patches going the other, and this would be adding a new kind of message that would need to be handled?

ept:
This PR is based on the Automerge 0.x release series. As we are now focussing our attention on the performance branch and the 1.x series, we can use the implementations of cursors in #307 or #313 instead, so I will close this PR. Thanks for getting us started on cursors, though!

PR #307, opened 2021-01-26
Initial implementation of cursors (perf branch version)

ept:
This PR is the same as #306, but based on the performance branch rather than main. It provides a first attempt at an API for cursors.

A cursor is an object of the form `{objectId: 'string', elemId: 'string'}` and it can be stored anywhere: inside an Automerge document or somewhere external. To get the cursor for a particular index of a text object, use:

```js
const cursor = doc.text.getCursorAt(index)
```

The cursor identifies a particular character in the text; the position of that character may change as text is inserted or deleted before it. To get the current index of a previously created cursor, use:

```js
const index = Automerge.getCursorIndex(doc, cursor, true)
```

The third argument controls what to do if the character referenced by the cursor has been deleted. If `false` and the character has been deleted, `getCursorIndex` returns -1. If the argument is `true` and the character has been deleted, `getCursorIndex` returns the index of the closest preceding non-deleted character, or -1 if there is no such preceding character.

I am not super happy with this approach, but putting up the PR nevertheless so that folks can play with it.

### DIFF
--- a/@types/automerge/index.d.ts
+++ b/@types/automerge/index.d.ts
@@ -42,6 +42,7 @@ declare module 'automerge' {
   function getAllChanges<T>(doc: Doc<T>): Uint8Array[]
   function getChanges<T>(olddoc: Doc<T>, newdoc: Doc<T>): Uint8Array[]
   function getConflicts<T>(doc: Doc<T>, key: keyof T): any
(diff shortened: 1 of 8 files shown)


ept:
I think we have come to the conclusion that the approach to cursors in #313 is preferable to this one, since it keeps open the option of having frontend and backend on separate threads. I will therefore close this PR.

PR #313, opened 2021-02-03
Alternative implementation of cursors

ept:
This PR is work in progress.

The implementation of cursors on #307 requires a direct call from the frontend into the backend, which is inconvenient when frontend and backend are on different threads. This PR is an alternative implementation that is integrated with the existing frontend-backend protocol, making it compatible with multiple threads.

More detail to come.

### DIFF
--- a/backend/columnar.js
+++ b/backend/columnar.js
@@ -29,7 +29,8 @@ const COLUMN_TYPE = {
 
 const VALUE_TYPE = {
   NULL: 0, FALSE: 1, TRUE: 2, LEB128_UINT: 3, LEB128_INT: 4, IEEE754: 5,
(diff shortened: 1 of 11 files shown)


ept:
This PR is now mostly complete and working: cursors can be created and stored, and indexes are automatically recomputed as needed. There is one known issue: `Automerge.load(Automerge.store(doc))` currently throws an exception when loading a document that contains cursors (there's a skipped test for this). Storing and loading a document should work if it is done as `Automerge.applyChanges(Automerge.init(), Automerge.getAllChanges(doc))`. I will try to fix this soonish, but otherwise this feature is ready for people to start playing with.

PR #322, opened 2021-02-24
Update deps and switch from Browserify to Webpack

heyheyhello:
Re: https://github.com/automerge/automerge/pull/321

I only have Firefox/Linux here and can't check your Saucelabs tests but this might help the polyfill issue. At least bundling is now done consistently for build+test. I'm not endorsing webpack here; it still takes 11s to build while esbuild takes 0.4s. Works though. The bundle analysis report should hopefully provide some insight into where automerge is heavy.

I wanted to remove _karma-typescript_ since I know webpack can use _ts-loader_ to process the one TS file you have. Unfortunately I couldn't get the path/module resolution to work no matter what I tried. All of my projects are TS, but I've never mixed JS/TS before so I think that's what's confusing me...


ept:
Thank you! It would be nice to remove karma-typescript at some point, but this PR is already good progress.

PR #324, opened 2021-03-05
Implementation of sync protocol

ept:
Here is a first attempt at an implementation of the sync protocol discussed in #290. Its purpose is to bring a local and a remote document into the same state. This is typically used when two peers have been disconnected for some time, and need to exchange any changes that happened while they were disconnected. The two peers that are syncing could be client and server, or server and client, or two peers with symmetric roles.

I am not totally happy with the API, but I wanted to share what I have anyway, to get some more eyes on it. You use it like this:

```js
// When a connection is established, one of the two peers sends the first message.
// For example, if using WebSocket, you could say the client sends the first message.
onConnect(() => {
  sync = Automerge.startSync(doc)
  sendMessage(sync.messageToSend)
})

onReceiveMessage((msg) => {
  const response = sync.processMessage(msg)
  if (response) sendMessage(response)
  if (sync.isFinished) {
    doc = Automerge.finishSync(doc, sync)
    lastSyncState = Automerge.getCurrentVersion(doc)
  }
})
```

The protocol is stateful, and the `sync` object returned by `Automerge.startSync()` manages this state. The document is not updated directly during the sync process, because the changes don't necessarily arrive all at once, and it would be annoying to get several half-updated states over the course of a sync. That's why only when `sync.isFinished` is true, we update the document based on the changes we received by calling `Automerge.finishSync(doc, sync)`. If a sync never reaches the point where `sync.isFinished` is true (perhaps the connection is interrupted and times out) then the document is not updated, and we start afresh with a new sync when reconnecting.

The intention is that we perform the sync only once when a connection is established; thereafter each peer can immediately send any new changes over the connection when they are made.

As implemented, the protocol requires both peers to load their documents from disk in order to sync them. However, there is an easy optimisation that saves loading any documents that have not changed. There is now a new function `Automerge.getCurrentVersion(doc)` that returns a short byte array that uniquely identifies the state of a document (it contains the hashes of the most recent changes); any two peers that are in the same state will return the same value for `getCurrentVersion`. You can store this current version separately from the document. When two peers, each with a large number of documents, want to find out which documents have changed, they only need to send each other this version identifier for each document; any docs whose version is the same need not be loaded from disk since they are unchanged. The actual sync protocol is then started only for those docs where the versions do not match.

The protocol is based on this paper: Martin Kleppmann and Heidi Howard. [Byzantine Eventual
Consistency and the Fundamental Limits of Peer-to-Peer Databases](https://arxiv.org/abs/2012.00472). arXiv:2012.00472, December 2020.

The protocol assumes that every time a peer successfully syncs with another peer, it remembers the document version (as returned by `Automerge.getCurrentVersion()`) after the last sync with that peer. The next time we try to sync with the same peer, we start from the assumption that the other peer's document version is no older than the outcome of the last sync, so we only need to exchange any changes that are more recent than the last sync. This assumption may not be true if the other peer did not correctly persist its state (perhaps it crashed before writing the result of the last sync to disk), and we fall back to sending the entire document in this case.

Let's say we want to sync A and B, and A is the first peer to send a message. Then the protocol state at A is initialised using the document version from the last sync with B (or `undefined` if this is the first time we're syncing with B). Then the peers exchange the following messages:

1. A sends B a message of type `MESSAGE_TYPE_BLOOM_REQ` containing the head hashes from the last sync between A and B, A's current head hashes, and a Bloom filter summarising all the changes that A knows about that were added since the last sync.
2. B responds to A with a message of type `MESSAGE_TYPE_BLOOM_RESP` containing the hashes of any changes that B wants from A, B's latest head hashes, a Bloom filter summarising all the changes that B knows about that were added since the last sync, and the changes that A needs from B (any changes that B has, that were added since the last sync and that are not contained in the Bloom filter; this set may be incomplete due to Bloom filter false positives).
3. If the last message contained at least one change or at least one hash that B wants, then A responds to B with a message of type `MESSAGE_TYPE_NEEDS`. This message contains the hashes of any changes that A wants from B, and any changes requested by B in its last message. This message is sent even if A does not want any more changes and B did not request any changes, in order to indicate that sync is complete.
4. If the last message indicated that A wanted further changes from B, then B replies with another `MESSAGE_TYPE_NEEDS` message containing those changes. Moreover, if there are any remaining hashes for which B does not have the corresponding changes, B also requests those in the same message. This process continues with further `MESSAGE_TYPE_NEEDS` messages (and A and B swapping roles each time) until both peers have all the changes (and their dependencies) contained in the other peer's head hashes in their initial message. The final `MESSAGE_TYPE_NEEDS` message is one containing no changes and no requested changes, which signals that the sync is complete (this message may be from A to B or from B to A).

Many documents can be synced over the same connection; you just need to tag each message with the ID of the document that it refers to, so that the recipient can deliver each message to the appropriate sync object.

Still need to add more tests to cover all the various edge cases.

The current implementation is likely to be quite slow. For example, when deciding whether to send a log of changes or a whole compressed document, it simply encodes both and then sends the one that is smaller. That will result in smaller messages over the network, but some wasted effort with encoding. Also, our current functions for traversing the hash graph are quite rudimentary. It would be good to figure out faster data structures for this stuff.

### DIFF
--- a/@types/automerge/index.d.ts
+++ b/@types/automerge/index.d.ts
@@ -59,13 +59,22 @@ declare module 'automerge' {
   function getAllChanges<T>(doc: Doc<T>): Uint8Array[]
   function getChanges<T>(olddoc: Doc<T>, newdoc: Doc<T>): Uint8Array[]
   function getConflicts<T>(doc: Doc<T>, key: keyof T): any
+  function getCurrentVersion<T>(doc: Doc<T>): Uint8Array
   function getHistory<D, T = Proxy<D>>(doc: Doc<T>): State<T>[]
   function getMissingDeps<T>(doc: Doc<T>): Hash[]
   function getObjectById<T>(doc: Doc<T>, objectId: OpId): any
   function getObjectId(object: any): OpId
 
+  function startSync<T>(doc: Doc<T>, initState?: Uint8Array): Sync
+  function finishSync<T>(doc: Doc<T>, sync: Sync, options?: ChangeOptions<T>): Doc<T>
   function load<T>(data: Uint8Array, options?: any): Doc<T>
   function save<T>(doc: Doc<T>): Uint8Array
+
+  class Sync {
+    readonly messageToSend: Uint8Array | undefined
+    readonly isFinished: boolean
+    processMessage(message: Uint8Array): Uint8Array | undefined
+  }
 
   // custom CRDT types
 
@@ -141,6 +150,7 @@ declare module 'automerge' {
     function free(state: BackendState): void
     function getAllChanges(state: BackendState): Uint8Array[]
     function getChanges(state: BackendState, haveDeps: Hash[]): Uint8Array[]
+    function getCurrentVersion(state: BackendState): Uint8Array[]
     function getHeads(state: BackendState): Hash[]
     function getMissingDeps(state: BackendState): Hash[]
     function getPatch(state: BackendState): Patch
@@ -148,6 +158,7 @@ declare module 'automerge' {
     function load(data: Uint8Array): BackendState
     function loadChanges(state: BackendState, changes: Uint8Array[]): BackendState
     function save(state: BackendState): Uint8Array
+    function startSync(state: BackendState, initState?: Uint8Array): Sync
   }
 
   // Internals
--- a/backend/index.js
+++ b/backend/index.js
@@ -3,6 +3,7 @@ const { copyObject } = require('../src/common')
 const OpSet = require('./op_set')
 const { SkipList } = require('./skip_list')
 const { splitContainers, encodeChange, decodeChanges, encodeDocument, constructPatch, BackendDoc } = require('./columnar')
+const { encodeCurrentVersion, Sync } = require('./sync')
 
 // Feature flag: false uses old Immutable.js-based backend data structures, true uses new
 // byte-array-based data structures. New data structures are not yet fully working.
@@ -245,10 +246,41 @@ function getHeads(backend) {
   return backend.heads
 }
 
+/**
+ * Returns a byte array that uniquely identifies the current state of the document.
+ * Any two documents that have applied the same set of changes (even in a different order)
+ * will return the same value from this function.
+ */
+function getCurrentVersion(backend) {
+  return encodeCurrentVersion(backend.heads)
+}
+
+/**
+ * Establishes a sync session with another peer. If the local peer is the initiator
+ * of the session, `initState` should be the document version (as returned by
+ * `getCurrentVersion`) of the last time we synced with that particular peer, or
+ * undefined if this is the first time we are syncing with this peer. If the
+ * remote peer initiated the session by sending us a message, `initState` is this
+ * message received from the other peer.
+ *
+ * Returns a `Sync` object that handles sending and receiving messages.
+ */
+function startSync(backend, initState) {
+  return new Sync(backendState(backend).get('opSet'), initState)
+}
+
+/**
+ * Returns the full history of changes that have been applied to a document.
+ */
 function getAllChanges(backend) {
   return getChanges(backend, [])
 }
 
+/**
+ * Returns all changes that are newer than or concurrent to the changes
+ * identified by the hashes in `haveDeps`. If `haveDeps` is an empty array, all
+ * changes are returned. Throws an exception if any of the given hashes is unknown.
+ */
 function getChanges(backend, haveDeps) {
   if (!Array.isArray(haveDeps)) {
     throw new TypeError('Pass an array of hashes to Backend.getChanges()')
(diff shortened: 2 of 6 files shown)


bionicles:
is it possible to implement sharded sync, for example, to have a central multitenant datastore, then only allow certain folks to sync certain subsets of the data, instead of the whole document?

also, can we make a recursive proxy sort of thing for the document, so we can just do crdt.nested.array[1] = "bob" 

see here https://www.youtube.com/watch?v=_5X2aB_mNp4

pvh:
@orionz, @alexjg and I spent some time workshopping the protocol and trying to make it more functional / stateless and came to this design:

---------
Four new functions:
- bloom = Automerge.Sync.createBloomFilter(doc)
- changeList = Automerge.Sync.estimateMissingChanges(doc, bloom)
- Automerge.Sync.{en,de}codeNetworkMessage()
These together allow a developer to implement a network protocol that fits their needs something like this:

-------
// on both peers
peer.onConnect() { 
  // calculating the bloom filter is a blocking operation
  var bloom = Automerge.Sync.createBloomFilter(doc)
  peer.send(Automerge.encodeNetworkMessage({BLOOM_FILTER: bloom}))
  peer.onMessage(Automerge.decodeNetworkMessage(msg)) {
    switch(msg) {
      BLOOM:
        var estimatedChanges = Automerge.Sync.estimateMissingChanges(doc, bloom)
        // always send changes oldest to newest in case of interruptions
        peer.send(Automerge.encodeNetworkMessage({CHANGES: estimatedChanges}))
      CHANGES:
        doc = Automerge.applyChanges(doc, msg.changes)
        // adding refuse prevents constantly re-requesting a change both peers don't have
        var missing = Automerge.getMissingChanges(doc) - msg.refused
        // we want to be careful that if we miss a change we don't desync
        // so getMissingChanges might require further optimization
        if (missing.empty) {
          // deliver document to frontend / renderer process
          // (we are synced)
        } else {
          peer.send(Automerge.encodeNetworkMessage({MISSING: missingChanges}))
        }
      MISSING:
        var changes = Automerge.getChanges(doc, msg.changes)
        peer.send(Automerge.encodeNetworkMessage(
          {CHANGES: changes, REFUSED: msg.changes - changes}
          )
        )
    }
  }
  
  // send ongoing local changes as they happen
  doc.onChange(change) {
    peer.send({CHANGES: [change]})
  }
}

pvh:
We also spent some time thinking about the "syncing with a filestore" case which seems very promising. Because the head hashes are at the start of the document, one can quickly confirm that the file in question (or range of a file!) contains only data we already have and either drop the connection or skip ahead to the next segment of the file.

That looks roughly like this:
```js
read() {
  offset = 0;
  var { sizeOfBlock, docHeads } 
  if (docHeads.any((head) !Automerge.alreadyContains(doc, head)) {
     // fetch this segment and apply it
  }
  offset += sizeOfBlock
}

```

HerbCaudill:
As a new convert to the Church of the State Machine, I'd argue that a sync protocol like this is inherently stateful - it's just a question of whether the library manages the state for you, or you have to do it yourself. 

This is the sort of thing that's really easy to get wrong, and as a developer I'd prefer to work with something that encapsulates the tricky stuff. 

It seems to me like the only thing I need control over is the network interface. I thought the API of the original Connection class made a lot of sense - give it your document and a function to send messages, and pass on any received messages; and it takes care of the rest. 

Of course there's no reason why we couldn't do both - expose the low-level parts, as well as something that puts it all together. If nothing else the higher-level thing can serve as an example for anyone who wants to DIY.

gterzian:
Regarding the API sketched at https://github.com/automerge/automerge/pull/324#issue-585699658, have you considered making some of the interface of the "sync" object event based? 

For example:

```js
onConnect(() => {
  sync = Automerge.startSync(doc)
  
  // Assume a web environment, 
  // using https://developer.mozilla.org/en-US/docs/Web/API/CustomEvent/CustomEvent
  sync.addEventListener('sendMessage', function (e) { sendMessage(e.detail.messageToSend) }, false);
  sync.addEventListener('finished', function (e) { 
      doc = Automerge.finishSync(doc, e.detail.syncResult) 
      lastSyncState = Automerge.getCurrentVersion(doc)
  }, false);
})

onReceiveMessage((msg) => {
  sync.processMessage(msg)
})
```

> The current implementation is likely to be quite slow.

I think the main advantage would be that all the sync logic could run in parallel to the main thread(for example, `sync.processMessage` could actually under the hood send a message to a worker), and communicate with the main thread via events. 

In the initial implementation, one could get the parallelism using some user-land concept like web workers(or the NodeJS equivalent), and one could also imagine Automerge one day being standardized, and the sync logic run by the engine in parallel to user code(using concepts as described at https://html.spec.whatwg.org/multipage/#event-loop-for-spec-authors).

ept:
I have made a revised API and implementation of this protocol and put it up on a new PR, #332. Feedback welcome.

@HerbCaudill I agree about the importance of state machines, but @pvh convinced me that the state machine should be under the control of the networking layer, not embedded in Automerge core, since different networking layers may have different requirements depending on the characteristics of the network they use. Certainly the protocol state machine isn't something that app developers should have to worry about — this is complexity that should be encapsulated in a library. The tests in #332 include an example of such a state machine, which can be used as a starting point for actual network protocol implementations.

@gterzian Event-based APIs have a lot of state, which can make them hard to reason about when there are lots of things happening concurrently. For this reason we have moved to a stateless, functional API for the core protocol primitives. It's always possible to make an event-based wrapper on top of it.

PR #328, opened 2021-03-12
Implement the new sequence patch format, fixes #311

### LINKED ISSUE #311: More compact change and patch format for list/text insertions/deletions
The current frontend-backend protocol is very verbose for insertions and deletions of sequences of list items (or characters in text). For example, creating the list `[1, 2, 3]` results in the following change:

```js
{
  "actor": "", "seq": 1, "startOp": 1, "time": 0, "deps": [], "ops": [
    {
      "action": "makeList",
      "obj": "_root",
      "key": "list",
      "pred": []
    },
    {
      "action": "set",
      "obj": "1@48b7977de26e44b29f35a1dc75c66d3c",
      "elemId": "_head",
      "insert": true,
      "value": 1,
      "pred": []
    },
    {
      "action": "set",
      "obj": "1@48b7977de26e44b29f35a1dc75c66d3c",
      "elemId": "2@48b7977de26e44b29f35a1dc75c66d3c",
      "insert": true,
      "value": 2
      "pred": []
    },
    {
      "action": "set",
      "obj": "1@48b7977de26e44b29f35a1dc75c66d3c",
      "elemId": "3@48b7977de26e44b29f35a1dc75c66d3c",
      "insert": true,
      "value": 3
      "pred": []
    }
}
```

and the following patch:

```js
{
  "objectId": "1@48b7977de26e44b29f35a1dc75c66d3c",
  "type": "list",
  "edits": [
    {"action": "insert", "index": 0, "elemId": "2@48b7977de26e44b29f35a1dc75c66d3c"},
    {"action": "insert", "index": 1, "elemId": "3@48b7977de26e44b29f35a1dc75c66d3c"},
    {"action": "insert", "index": 2, "elemId": "4@48b7977de26e44b29f35a1dc75c66d3c"}
  ],
  "props": {
    "0": {"2@48b7977de26e44b29f35a1dc75c66d3c": {"value": 1}},
    "1": {"3@48b7977de26e44b29f35a1dc75c66d3c": {"value": 2}},
    "2": {"4@48b7977de26e44b29f35a1dc75c66d3c": {"value": 3}}
  }
}
```

Creating an Automerge.Text object with the content `'a', 'b', 'c'` looks very similar. If you copy and paste a few kilobytes of text, the JSON is huge.

This verbosity is not only an efficiency issue, but the patch format is also painful for applications to work with when using the observable API (#308). Note that the indexes of elements to insert are in `edits`, while the values associated with those indexes are in `props`. This seemed like a good idea at the time, but it's pretty annoying now…

I propose extending the format for changes and patches to group together consecutive insertions or deletions. This affects only the JSON format used in the frontend-backend protocol; the binary format already uses compression that performs this grouping. For example, the above change might be represented as:

```js
{
  "actor": "", "seq": 1, "startOp": 1, "time": 0, "deps": [], "ops": [
    {
      "action": "makeList",
      "obj": "_root",
      "key": "list",
      "pred": []
    },
    {
      "action": "set",
      "obj": "1@48b7977de26e44b29f35a1dc75c66d3c",
      "elemId": "_head",
      "insert": true,
      "values": [1, 2, 3]
    }
}
```

For an insertion of multiple consecutive values into a list/text, an operation with a `values` array expands into multiple operations, one per element of the array. To keep things simple, I propose that this optimisation is only used when all of the values all primitive datatypes representable in JSON (string, boolean, number, null). For nested objects we use a `make*` operation like we do now, and for values not representable in JSON (e.g. Date or Counter objects) we use a `set` operation with a single value and a datatype tag, like we do now.

We can also optimise the representation of sequential deletions, assuming that the elemIds of the elements being deleted all have the same actor, and consecutive counter values. For example, to delete the three elements we inserted earlier:

```js
    {
      "action": "del",
      "obj": "1@48b7977de26e44b29f35a1dc75c66d3c",
      "elemId": "2@48b7977de26e44b29f35a1dc75c66d3c",
      "multiOp": 3
    }
```

The `multiOp` property indicates that this deletion should be expanded into three deletions, deleting elements with IDs `2@48b7977de26e44b29f35a1dc75c66d3c`, `3@48b7977de26e44b29f35a1dc75c66d3c`, and `4@48b7977de26e44b29f35a1dc75c66d3c` respectively. If we want to delete a range of elements whose elemIds contain more than one actor, or whose counters are not consecutive, we first break the deleted range into sub-ranges that are consecutive, and issue a deletion op for each sub-range.

In the opposite direction (backend to frontend), we can change the patch format to:

```js
{
  "objectId": "1@48b7977de26e44b29f35a1dc75c66d3c",
  "type": "list",
  "edits": [
    {"action": "insert", "index": 0, "elemId": "2@48b7977de26e44b29f35a1dc75c66d3c", "values": [1, 2, 3]}
  ]
}
```

Again this works with primitive values; I'm less clear on how to represent nested objects, datatypes that need a type tag (e.g. Date), and conflicts (multiple values associated with the same list element). For consecutive deletions, the patch only needs to contain the start index and number of elements to delete.

If we're putting all the inserted values in the `edits` section of the patch, then what about assigning the value of an existing list element? At the moment assignments are done through the `props` section, but maybe they should go in the `edits` array as well.

For text insertions, a further thing we could do: rather than making the `values` property an array of individual characters, we could just put the raw string (all the characters concatenated) directly in the change/patch, and have the backend break it apart into individual characters. However, this does raise the question of how exactly we should handle complex unicode characters. Perhaps the simplest decent solution is to [use `Array.from(str)`](https://stackoverflow.com/questions/4547609/how-to-get-character-array-from-a-string) to break a string into Unicode code points, although this does mess up characters that are a combination of several code points (e.g. accents, or some emoji). A more thorough approach would be to [split the string into grapheme clusters](https://github.com/orling/grapheme-splitter), but that's getting beyond the scope of this issue.

Any thoughts or concerns? Obviously this change will need to be coordinated with automerge-rs.

alexjg:
This is a first shot at implementing the new sequence patch format outlined in #311. It's mostly working with the exception of the Observable API. I need to spend some time figuring that out but I wanted to get some eyes on this first.

There is one outstanding question I have which relates to the value of the pred attribute for multiOp delete operations. Consider this operation:

```javascript
{
  obj: listId, 
  action: 'del', 
  elemId: '1@xxx', 
  multiOp: 2,
  insert: false,
}
```

In order for this to work we need to specify the predecessor operations for each element being overwritten. The approach I've gone for in this patch is to flatten the predecessor IDs for the whole set of deleted list elements into one pred array and attached it to the op. This works but it feels clumsy and belies the intent of this protocol change - which is to reduce the size of these operations - as we have to include at least one predecessor ID for every deleted element. Is there a better way? Have I misunderstood the use of the pred attribute in the protocol?

### DIFF
--- a/@types/automerge/index.d.ts
+++ b/@types/automerge/index.d.ts
@@ -266,7 +266,8 @@ declare module 'automerge' {
     action: 'insert'
     index: number   // the list index at which to insert the new element
     elemId: OpId    // the unique element ID of the new list element
-    value: MapDiff | ListDiff | ValueDiff 
+    opId: OpId      // ID of the operation that assigned this value
+    value: MapDiff | ListDiff | ValueDiff
   }
 
   // Describes the insertion of a consecutive sequence of primitive values into
@@ -289,8 +290,8 @@ declare module 'automerge' {
   interface UpdateEdit {
     action: 'update'
     index: number   // the list index to update
-    opId: OpId      // ID of the operation that performed this update
-    value: MapDiff | ListDiff | ValueDiff 
+    opId: OpId      // ID of the operation that assigned this value
+    value: MapDiff | ListDiff | ValueDiff
   }
 
   // Describes the deletion of one or more consecutive elements from a list or
--- a/@types/automerge/index.d.ts
+++ b/@types/automerge/index.d.ts
@@ -39,7 +39,7 @@ declare module 'automerge' {
     }
 
   type PatchCallback<T> = (patch: Patch, before: T, after: T, local: boolean, changes: BinaryChange[]) => void
-  type ObserverCallback<T> = (diff: MapDiff | ListDiff | ValueDiff , before: T, after: T, local: boolean, changes: BinaryChange[]) => void
+  type ObserverCallback<T> = (diff: MapDiff | ListDiff | ValueDiff, before: T, after: T, local: boolean, changes: BinaryChange[]) => void
 
   class Observable {
     observe<T>(object: T, callback: ObserverCallback<T>): void
--- a/@types/automerge/index.d.ts
+++ b/@types/automerge/index.d.ts
@@ -39,7 +39,7 @@ declare module 'automerge' {
     }
 
   type PatchCallback<T> = (patch: Patch, before: T, after: T, local: boolean, changes: BinaryChange[]) => void
-  type ObserverCallback<T> = (diff: ObjectDiff, before: T, after: T, local: boolean, changes: BinaryChange[]) => void
+  type ObserverCallback<T> = (diff: MapDiff | ListDiff | ValueDiff , before: T, after: T, local: boolean, changes: BinaryChange[]) => void
 
   class Observable {
     observe<T>(object: T, callback: ObserverCallback<T>): void
@@ -226,6 +226,8 @@ declare module 'automerge' {
     value?: number | boolean | string | null
     datatype?: DataType
     pred?: OpId[]
+    values?: (number | boolean | string | null)[]
+    multiOp?: number
   }
 
   interface Patch {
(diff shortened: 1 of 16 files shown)


ept:
Hi @alexjg, I'm now working on this (I have some work-in-progress commits on this branch that I've not yet pushed). Could you hold off on force-pushing it please, since it means I have to rebase my stuff too?

alexjg:
Ah apologies, I didn't realise you were working on it. I've just been keeping it up to date with the work being merged into `performance`.

ept:
No worries, that makes sense and I should have said I was working on it! Anyway, sorry for the long delay on this, and I will have more news soon.

ept:
Okay, I have finally finished with this. Sorry it has taken so long, and thank you @alexjg for doing the original work. I have found and fixed a number of bugs, and added tests for them. Most importantly, in order to group several insertions into a multi-insert they must all have sequential elemIds (i.e. the actor parts of the elemIds are all the same, and the counter part increments by 1 for each element), and in order to group several deletions they must have sequential elemIds and sequential preds. I think the code now ensures this.

I expect automerge/automerge-rs#88 will need updating correspondingly. @alexjg @orionz @jeffa5 could one of you help me with this please?

I am going to go ahead and merge this because folks are desperate to get a 1.0-preview release on npm; apologies for not waiting for the Rust implementation to reach parity. We can follow up with a release of the wasm backend when the equivalent of this PR is also finished on the Rust side.

Issue #317, opened 2021-02-11
Provide built-in general-purpose compression


ept:
Although our compressed binary format (#253) already uses some compression techniques such as RLE, I have found that applying an additional layer of gzip compression may yield an additional factor of 2 or so compression on the text-editing benchmark (for the whole-document encoding). I suspect that for many applications, it would make sense to use such an additional level of compression.

The question is: should this compression be built in to Automerge itself, or should it be handled in a separate I/O layer (when reading/writing documents on disk, when transferring documents over a network)? There are precedents for both:

* On the web, HTML, CSS and JS files are typically stored and manipulated uncompressed, and they are compressed on-the-fly when sent over the network (as indicated e.g. by the HTTP header `Transfer-Encoding: gzip`).
* On the other hand, PNG and JPG files have built-in compression, and they are both stored and transferred in compressed form. The libraries that encode/decode these formats include the compression algorithm out of the box. Applications never normally see the decompressed byte stream.

I am beginning to think that maybe the image file model would be a better approach for Automerge, since the binary encoding already requires a dedicated decoder anyway, and having all the compression built-in would simplify the job for users who want to integrate Automerge with various storage and networking libraries. The downside is that we would have to bundle a compression codec with Automerge, increasing the size of the library for any users who don't need it. But I suspect it would be worth turning such compression on by default.

If we do this, we will have to decide on a compression algorithm to use: [DEFLATE](https://tools.ietf.org/html/rfc1951), [LZ4](https://github.com/lz4/lz4), [Snappy](https://github.com/google/snappy), [bzip2](https://en.wikipedia.org/wiki/Bzip2), [Zstandard](https://github.com/facebook/zstd), or something else? There is a bit of a speed-size trade-off here. Whatever algorithm we use, we should pick one that has good implementations in both JS and Rust. Adopting a widely-used algorithm would make it easier for people who want to write their own tooling around the Automerge data format.

Any thoughts?

savaki:
I like it.  If we follow parquet's lead, then being able to choose the compression on a per column basis could be really useful.  Columns like value could benefit highly from this whereas columns like insert might not see that much of a gain.

My vote would be for LZ4 for fast decompression speeds.

HerbCaudill:
> I am beginning to think that maybe the image file model would be a better approach for Automerge, since the binary encoding already requires a dedicated decoder anyway

Agree - the HTML/CSS analogy might have held when the contents were somewhat human-readable, but not with the new encoding.

[...]

PR #330, opened 2021-03-16
Add DEFLATE compression

ept:
This PR adds support for DEFLATE compression. As explained in #317, compression can be used in two places:

* On an individual change, the whole change can be compressed. The change hash is computed on the uncompressed change, and the checksum included in the change is based on the uncompressed change. We do this because two different compressors given the same input do not necessarily produce the same compressed byte stream.
* On a compressed document, one column at a time can be compressed. I changed the definition of the columnId to reserve one bit to indicate whether the column is compressed. We do this so that it is possible to load only a subset of columns of a document.

Compression is only enabled when the data is larger than some threshold (currently 256 bytes) since it's not worth it on very short values.

@alexjg could I leave the implementation on the Rust side with you?

### DIFF
--- a/backend/columnar.js
+++ b/backend/columnar.js
@@ -1,3 +1,4 @@
+const pako = require('pako')
 const { copyObject, parseOpId, equalBytes } = require('../src/common')
 const {
   utf8ToString, hexStringToBytes, bytesToHexString,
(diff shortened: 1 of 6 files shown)


alexjg:
Looks pretty straightforward, I'll pick it up for the Rust implementation later this week.

alexjg:
The Rust implementation of this is ready to merge now.

ept:
@alexjg Excellent, thank you! I have added some JS-wasm interop tests to check that deflate-compressed changes and documents work correctly in both directions. Merging this into `performance` now.

PR #334, opened 2021-04-01
Support creating text with unicode chars

nemanja-tosic:
Creating a text with a unicode character results in the unicode characted being split. If used with the rust backend it results in an ugly runtime error.

See: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/split

There is a warning in there:

> Warning: When the empty string ("") is used as a separator, the string is not split by user-perceived characters (grapheme clusters) or unicode characters (codepoints), but by UTF-16 codeunits. This destroys surrogate pairs. See “How do you get a string to a character array in JavaScript?” on StackOverflow.

I did not do an extensive check of the codebase for this issue, just fixed the most critical path for us at this point.

### DIFF
--- a/frontend/text.js
+++ b/frontend/text.js
@@ -4,7 +4,7 @@ const { isObject } = require('../src/common')
 class Text {
   constructor (text) {
     if (typeof text === 'string') {
-      const elems = text.split('').map(value => ({value}))
+      const elems = [...text].map(value => ({value}))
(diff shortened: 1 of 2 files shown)


ept:
I've meant to fix this but keep forgetting. Thank you for the contribution!

PR #340, opened 2021-04-22
Add fast path to OpSet.getMissingChanges

ept:
Improves performance in the common case where user wants to just get the most recent couple of changes, e.g. in the context of a frequent sync. In this case, the cost becomes linear in the number of changes returned, whereas previously it was linear in the size of the entire change history.

In [this benchmark](https://gist.github.com/pvh/73473686865b9e421b620bd308fc9c58), the case of 1,000 changes and syncing after each change speeds up by roughly 10x after applying this patch.

### DIFF
--- a/backend/op_set.js
+++ b/backend/op_set.js
@@ -485,7 +485,43 @@ function getChangeByHash(opSet, hash) {
  * if any of the given hashes are not known to this replica.
  */
 function getMissingChanges(opSet, haveDeps) {
-  let stack = haveDeps, seenHashes = {}
+  // If the other replica has nothing, return all changes in history order
+  if (haveDeps.isEmpty()) {
+    return opSet.get('history').toJSON().map(hash => getChangeByHash(opSet, hash))
+  }
+
+  // Fast path for the common case where all new changes depend only on haveDeps
+  let stack = List(), seenHashes = {}, toReturn = []
+  for (let hash of haveDeps) {
+    seenHashes[hash] = true
+    const successors = opSet.getIn(['hashes', hash, 'depsFuture'])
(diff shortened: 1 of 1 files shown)


backend/op_set.js
```diff
@@ -485,7 +485,43 @@ function getChangeByHash(opSet, hash) {
...
+
+  // Depth-first traversal of the hash graph to find all changes that depend on `haveDeps`
+  while (!stack.isEmpty()) {
+    const hash = stack.last()
+    seenHashes[hash] = true
+    toReturn.push(hash)
+    if (!opSet.getIn(['hashes', hash, 'depsPast']).every(dep => seenHashes[dep])) {
+      // If a change depends on a hash we have not seen, abort the traversal and fall back to the
+      // slower algorithm. This will sometimes abort even if all new changes depend on `haveDeps`,
+      // because our depth-first traversal is not necessarily a topological sort of the graph.
```
jeffa5, line 510: I like this optimisation for the cases when we're syncing a lot, regarding the comment of topological sort have we thought about keeping a topological sort in the backend and incrementally adding to it? I'd then expect the hashes map to keep indexes into that. I think this would be useful for this case though not sure how wide reaching it would be?

ept: The `history` key contains a topologically sorted list of hashes — not any particular one, just the order in which we applied changes. That's why a few lines down we do `opSet.get('history').filter(...).map(...)` to get changes in topologically sorted order.

Mapping hashes to indexes in that list is a good idea. That would allow us to jump to the appropriate position and we'd just need to scan forward from there. I'll try implementing that later.

jeffa5: Ah yes, the history is what I had in mind. Looking forward to it :)

PR #344, opened 2021-04-23
Store sentHashes as a set rather than a list

jeffa5:
This means that we don't have to rebuild it every time we need to
generate a sync message.


pvh (approved):

backend/sync.js
```diff
@@ -379,22 +379,21 @@ function generateSyncMessage(backend, syncState) {
...
   }
 
   // Regular response to a sync message: send any changes that the other node
   // doesn't have. We leave the "have" field empty because the previous message
   // generated by `syncStart` already indicated what changes we have.
   const syncMessage = {heads: ourHeads, have: ourHave, need: ourNeed, changes: changesToSend}
-  const hashesToSend = changesToSend.map(change => decodeChangeMeta(change, true).hash)
+  for (const change of changesToSend) {
+    sentHashes[decodeChangeMeta(change, true).hash] = true
+  }
```
ept, line 392: Note that this mutates the `sentHashes` object that is passed in. In previous conversations, @pvh was adamant that these functions should be totally side-effect-free.

ept: Fixed in 6468bafb01bbe313f831bce5ce789be878cc5459

PR #345, opened 2021-04-23
Enable a gentle eslint

pvh:
This is a draft patch enabling `eslint` for Automerge. This first patch leaves a number of known bugs and problems found by eslint in place in the interest of first making some decisions about whether we want to move forward with this approach or not.

In almost every case I've either disabled an eslint rule project-wide or on the particular line where the complaint from the linter occurred, but in a few cases I decided the fixes were uncontroversial enough just to tidy up on the spot.

In particular, the most important rule I think we still ought to enable is `no-unused-vars`. That will require making a decision about tests which have loads of unused variables that are kept around for visual consistency. I think it would be a fine compromise to enable no-unused-vars overall but consider leaving it disabled for tests (at least for the time being.)

I'll add some comments on the patch below. If we can reach an agreeable set of eslint rules and land them we can always tune them later, either adding new rules to improve consistency (many of the rules I disabled from the base were only triggered by one or two cases) or to better suit Martin's preferred style.

### DIFF
--- a/backend/columnar.js
+++ b/backend/columnar.js
@@ -516,7 +516,7 @@ function decoderByColumnId(columnId, buffer) {
 function makeDecoders(columns, columnSpec) {
   // By default, every column decodes an empty byte array
   const emptyBuf = Uint8Array.of(), decoders = {}
-  for (let [/*columnName*/, columnId] of Object.entries(columnSpec)) {
+  for (let [/* columnName */, columnId] of Object.entries(columnSpec)) {
     decoders[columnId] = decoderByColumnId(columnId, emptyBuf)
   }
   for (let column of columns) {
@@ -525,7 +525,7 @@ function makeDecoders(columns, columnSpec) {
 
   let result = []
   for (let columnId of Object.keys(decoders).map(id => parseInt(id)).sort((a, b) => a - b)) {
-    let [columnName] = Object.entries(columnSpec).find(([/*name*/, id]) => id === columnId)
+    let [columnName] = Object.entries(columnSpec).find(([/* name */, id]) => id === columnId)
     if (!columnName) columnName = columnId.toString()
     result.push({columnId, columnName, decoder: decoders[columnId]})
   }
@@ -1279,9 +1279,9 @@ function constructPatch(documentBuffer) {
  *     elements that precede the position where the new operations should be applied.
  */
 function seekToOp(ops, docCols, actorIds) {
-  const { objActor, objCtr, keyActor, keyCtr, keyStr, idActor, idCtr, insert /*, action, consecutiveOps*/ } = ops
+  const { objActor, objCtr, keyActor, keyCtr, keyStr, idActor, idCtr, insert /* , action, consecutiveOps */ } = ops
   const [objActorD, objCtrD, /* keyActorD */, /* keyCtrD */, keyStrD, idActorD, idCtrD, insertD, actionD,
-    /* valLenD */, /* valRawD */, /* chldActorD */, /*chldCtrD*/, succNumD] = docCols.map(col => col.decoder)
+    /* valLenD */, /* valRawD */, /* chldActorD */, /* chldCtrD */, succNumD] = docCols.map(col => col.decoder)
   let skipCount = 0, visibleCount = 0, elemVisible = false, nextObjActor = null, nextObjCtr = null
   let nextIdActor = null, nextIdCtr = null, nextKeyStr = null, nextInsert = null, nextSuccNum = 0
 
@@ -1594,7 +1594,7 @@ function appendOperation(outCols, inCols, operation) {
  */
 function groupRelatedOps(change, changeCols, objectMeta) {
   const currentActor = change.actorIds[0]
-  const [objActorD, objCtrD, keyActorD, keyCtrD, keyStrD, /*idActorD*/, /*idCtrD*/, insertD, actionD] =
+  const [objActorD, objCtrD, keyActorD, keyCtrD, keyStrD, /* idActorD */, /* idCtrD */, insertD, actionD] =
     changeCols.map(col => col.decoder)
   let objIdSeen = {}, firstOp = null, lastOp = null, opIdCtr = change.startOp
   let opSequences = [], objectIds = {}
--- a/backend/columnar.js
+++ b/backend/columnar.js
@@ -516,7 +516,7 @@ function decoderByColumnId(columnId, buffer) {
 function makeDecoders(columns, columnSpec) {
   // By default, every column decodes an empty byte array
   const emptyBuf = Uint8Array.of(), decoders = {}
-  for (let [/* columnName */, columnId] of Object.entries(columnSpec)) {
+  for (let columnId of Object.values(columnSpec)) {
     decoders[columnId] = decoderByColumnId(columnId, emptyBuf)
   }
   for (let column of columns) {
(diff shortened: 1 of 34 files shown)


backend/columnar.js
```diff
@@ -1736,7 +1736,7 @@ class BackendDoc {
       valLen = 9, valRaw = 10, predNum = 13, predActor = 14, predCtr = 15, succNum = 13, succActor = 14, succCtr = 15
 
     const objectId = ops.objId
-    const elemId = op[keyStr] ? op[keyStr] :
+    const elemId = op[keyStr] ? op[keyStr] : // eslint-disable-line
```
pvh, line 1739: There is a rule (which I disabled) which makes it an error to nest ternary operators because they're difficult to read & reason about. I think we probably ought to enable that, which would make the actual rule (something about the colon being in an inconsistent place) irrelevant here.

backend/columnar.js
```diff
@@ -2091,9 +2091,7 @@ class BackendDoc {
     docState.lastIndex[ops.objId] = visibleCount
     for (let col of docState.opsCols) col.decoder.reset()
 
-    let outCols = docState.allCols.map(columnId => {
-      return {columnId, encoder: encoderByColumnId(columnId)}
-    })
+    let outCols = docState.allCols.map(columnId => ({columnId, encoder: encoderByColumnId(columnId)}))
```
pvh, line 2094: this comes up a few times -- if you have an arrow function returning an object you can just wrap it in parentheses instead of creating a block and returning the value

backend/skip_list.js
```diff
@@ -114,7 +114,7 @@ class SkipList {
   constructor (randomSource) {
     const head = new Node(null, null, 1, [], [null], [], [null])
     const random = randomSource ? randomSource() : randomLevel()
-    return makeInstance(0, Map().set(null, head), random)
+    return makeInstance(0, Map().set(null, head), random) // eslint-disable-line
```
pvh, line 117: I'm not sure the best way to fix this, but generally speaking a Javascript constructor doesn't return the object created but manipulates "this". I suspect a fix will be a little bit tricky.

backend/sync.js
```diff
@@ -440,7 +440,7 @@ function receiveSyncMessage(backend, oldSyncState, binaryMessage) {
   // changes without applying them. The set of changes may also be incomplete if the sender decided
   // to break a large set of changes into chunks.
   if (message.changes.length > 0) {
-    ;[backend, patch] = Backend.applyChanges(backend, message.changes)
+    [backend, patch] = Backend.applyChanges(backend, message.changes)
```
pvh, line 443: I elected to remove leading `;` from lines where it wasn't necessary, because eslint will detect when it is. These are the only semicolons we use throughout the codebase (at least that I found) but we use this idiom fairly often.

frontend/index.js
```diff
@@ -87,7 +87,7 @@ function makeChange(doc, context, options) {
     seq: state.seq,
     startOp: state.maxOp + 1,
     deps: state.deps,
-    time: (options && typeof options.time === 'number') ? options.time :
+    time: (options && typeof options.time === 'number') ? options.time : // eslint-disable-line
```
pvh, line 90: There is some variation about whether the colon appears on the following line or here. It looks like usually it's a colon on the next line.

frontend/proxies.js
```diff
@@ -89,7 +89,7 @@ function listMethods(context, listId, path) {
     methods[method] = (...args) => {
       const list = context.getObject(listId)
         .map((item, index) => context.getObjectField(path, listId, index))
-      return list[method].call(list, ...args)
+      return list[method].call(list, ...args) // eslint-disable-line
```
pvh, line 92: This is apparently an "unnecessary use of call" but I figured it should get a follow-up patch.

frontend/text.js
```diff
@@ -169,7 +169,7 @@ for (let method of ['concat', 'every', 'filter', 'find', 'findIndex', 'forEach',
                     'slice', 'some', 'toLocaleString']) {
   Text.prototype[method] = function (...args) {
     const array = [...this]
-    return array[method].call(array, ...args)
+    return array[method].call(array, ...args) // eslint-disable-line
```
pvh, line 172: These are more of the returns that appear to be non-standard. This is probably undesirable behaviour but is user-facing so we should consider a change more carefully (and try to fix it before 1.0 if a fix is required.)

test/frontend_test.js
```diff
@@ -86,7 +86,7 @@ describe('Automerge.Frontend', () => {
     it('should delete keys in maps', () => {
       const actor = uuid()
       const [doc1, change1] = Frontend.change(Frontend.init(actor), doc => { doc.magpies = 2; doc.sparrows = 15 })
-      const [doc2, change2] = Frontend.change(doc1, doc => delete doc['magpies'])
+      const [doc2, change2] = Frontend.change(doc1, doc => delete doc['magpies']) // eslint-disable-line
```
pvh, line 89: eslint encourages consistent use of dot notation where the property is precomputed, but these uses seemed deliberate. Thoughts on that, @ept?

test/fuzz_test.js
```diff
@@ -119,7 +119,7 @@ class Micromerge {
       if (!meta[index].deleted) visible++
       index++
     }
-    if (index === meta.length) throw new RangeError(`List element not found: ${op.key}`)
+    if (index === meta.length) throw new RangeError(`List element not found: ${op.key}`) /* this is a bug! */ // eslint-disable-line
```
pvh, line 122: op's not defined here -- looks leftover from an earlier version.

test/proxies_test.js
```diff
@@ -199,7 +200,7 @@ describe('Automerge proxy API', () => {
           assert.strictEqual(doc.list.every(val => val > 0), true)
           assert.strictEqual(doc.list.every(val => val > 2), false)
           assert.strictEqual(doc.list.every((val, index) => index < 3), true)
-          doc.list.every(function () { assert.strictEqual(this.hello, 'world') }, {hello: 'world'})
+          doc.list.every(function () { assert.strictEqual(this.hello, 'world') }, {hello: 'world'}) // eslint-disable-line
```
pvh, line 203: I really don't understand what the intent is with these calls. Is `this` changing on each iteration?

test/skip_list_test.js
```diff
@@ -192,11 +192,11 @@ describe('SkipList', () => {
 
     it('should behave like a JS array', () => {
       jsc.assert(jsc.forall(jsc.bless({generator: makeSkipListOps}), function (ops) {
-        let levels = ops.filter(op => op.hasOwnProperty('insertAfter')).map(op => op.level)
+        let levels = ops.filter(op => op.hasOwnProperty('insertAfter')).map(op => op.level) // eslint-disable-line
```
pvh, line 195: using hasOwnProperty on an object directly is problematic because if someone passes an object like this: `{ hasOwnProperty: "HAXXXOR" }` it will cause a crash. Instead the recommendation is to defensively call op.prototype.hasOwnProperty()

test/skip_list_test.js
```diff
@@ -205,6 +205,7 @@ describe('SkipList', () => {
           }
         }
 
+        // eslint-disable-next-line
```
pvh, line 208: Safe to delete this commented-out code?

pvh:
Okay, I've now cleaned up all the remaining easy stuff. There is a little bit of test weirdness that results in a few disabled rules for those test files but the actual project code is linting clean. I just saw in the Travis build above that some version of Edge doesn't like something in one of the tests but it's a bit late to start debugging Microsoft browsers over here, sorry -- hope it's either a CI glitch or else really obvious.

Other than that, the last bit that really remains here is fixing the one real bug, which is the way we're returning objects from constructors. This will likely result in an API change and so I didn't want to include it in this branch.

Let me know if you have questions / see any problems and I'll weigh in when I arise tomorrow.

test/proxies_test.js
ept: The intent is to check that when the callback function is called, `this` is set to the second argument passed to `every` (to match the behaviour of [Array.every](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/every)). I added a comment to explain.

test/skip_list_test.js
ept: 👍

ept:
I'm happy with this, and have added a few more small cleanups. Thanks @pvh!

PR #347, opened 2021-04-27
adds failing test

orionz:
Lukas found a issue and wrote a test for it - I think this may be a result of taking out the op coalescing frontend but we had no tests catching it.

Perhaps we need to put the op coalescing back in or make the backend properly handle multiple increments in the same change.


ept:
Marking this as needing to be fixed before 1.0 stable.

PR #348, opened 2021-04-27
Speedup hashesbyactor

orionz:
Was running a benchmark doing 1 char insert per change over 50,000 chars and it ran in 80 seconds - this change makes it run in 16 seconds.


ept:
Nice, thank you @orionz!

PR #350, opened 2021-04-30
Polymorphic insertAt for Text

### LINKED ISSUE #326: accept strings in Automerge.Text.insertAt
I sometimes have to do bulk inserts into an Automerge.Text, e.g. when importing a new document.

The current API accepts inserts one character at a time: `text.insertAt(0, 'h', 'e', 'l', 'l', 'o')`. Doing this with bulk inserts I've hit the javascript limit on the number of arguments allowed to a function call (65536).

I'd like it if insertAt also accepted strings and treated them indistinguishably from individual characters: `text.insertAt(0, 'hello')` (and maybe `text.insertAt(0, 'he', 'llo')`, not sure).

Or alternatively, a new API that accepted only one single string: `text.insertStringAt(0, 'hello')`.

I'm willing to attempt an implementation if you're open to it.

lightsprint09:
This could resolve #326. 

It introduces a the possibility to pass normal strings into `insertAt`.
```js
Automerge.change(s1, doc => doc.text.insertAt(0, 'string'))
```

 If one passes a string it gets split up into characters inside `insertAt` and processed as as list of single characters

### DIFF
--- a/frontend/text.js
+++ b/frontend/text.js
@@ -137,6 +137,9 @@ class Text {
    * Inserts new list items `values` starting at position `index`.
    */
   insertAt(index, ...values) {
+    if (typeof values[0] === 'string' && values[0].length > 1) {
+      values = values[0].split("")
+    }
     if (this.context) {
       this.context.splice(this.path, index, 0, values)
     } else if (!this[OBJECT_ID]) {
--- a/frontend/text.js
+++ b/frontend/text.js
@@ -138,7 +138,7 @@ class Text {
    */
   insertAt(index, ...values) {
     if (typeof values[0] === 'string' && values[0].length > 1) {
-      values = values[0].split("")
+      values = [...values[0]]
     }
     if (this.context) {
       this.context.splice(this.path, index, 0, values)
--- a/frontend/text.js
+++ b/frontend/text.js
@@ -134,16 +134,23 @@ class Text {
   }
 
   /**
-   * Inserts new list items `values` starting at position `index`.
+   * Inserts one or more new characters starting at position `index`. `characters` can either be a
+   * string (in which case it will be split into a sequence of Unicode code points), or an array of
+   * strings (in which case each element of the array is treated as a character and not split any
+   * further). The latter mode is useful if you want to use custom logic for splitting the string,
+   * for example to split it into Unicode grapheme clusters.
    */
-  insertAt(index, ...values) {
-    if (typeof values[0] === 'string' && values[0].length > 1) {
-      values = [...values[0]]
+  insertAt(index, characters) {
+    if (typeof characters === 'string') {
+      characters = [...characters]
+    } else if (!Array.isArray(characters)) {
+      throw new TypeError('Text.insertAt takes either a string or an array of characters')
     }
+
     if (this.context) {
-      this.context.splice(this.path, index, 0, values)
+      this.context.splice(this.path, index, 0, characters)
     } else if (!this[OBJECT_ID]) {
-      this.elems.splice(index, 0, ...values.map(value => ({value})))
+      this.elems.splice(index, 0, ...characters.map(value => ({value})))
     } else {
       throw new TypeError('Automerge.Text object cannot be modified outside of a change block')
     }
--- a/test/text_test.js
+++ b/test/text_test.js
@@ -135,7 +135,7 @@ function applyRetainOp(text, offset, op) {
   let length = op.retain
 
   if (op.attributes) {
-    text.insertAt(offset, { attributes: op.attributes })
+    text.insertAt(offset, [{ attributes: op.attributes }])
     offset += 1
   }
 
@@ -148,7 +148,7 @@ function applyRetainOp(text, offset, op) {
   }
 
   if (op.attributes) {
-    text.insertAt(offset, { attributes: inverseAttributes(op.attributes) })
+    text.insertAt(offset, [{ attributes: inverseAttributes(op.attributes) }])
     offset += 1
   }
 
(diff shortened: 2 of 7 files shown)


frontend/text.js
```diff
@@ -137,6 +137,9 @@ class Text {
    * Inserts new list items `values` starting at position `index`.
    */
   insertAt(index, ...values) {
+    if (typeof values[0] === 'string' && values[0].length > 1) {
+      values = values[0].split("")
```
nemanja-tosic, line 141: This destroys surrogate pairs. See #334. You should use `values = [...values[0]]`

test/text_test.js
```diff
@@ -201,14 +201,27 @@ describe('Automerge.Text', () => {
     s2 = Automerge.merge(Automerge.init(), s1)
   })
 
-  it('should support insertion', () => {
+  it('should support insertion of chacracter', () => {
```
nemanja-tosic, line 204:
```suggestion
  it('should support insertion of character', () => {
```

nemanja-tosic:
What is the expected behavior for `text.insertAt(0, 'abc')`?

One test specifying "should support inserting a single element" is asserting that the text equals abc: https://github.com/automerge/automerge/blob/performance/test/typescript_test.ts#L327-L331

This test is in contradiction with the PR and should fail, and is only passing because there is an assertion on the stringified value of text, which will be 'abc' in either case - single element or multiple chars.

Another test is asserting that the text object contains an with the value of abc: https://github.com/automerge/automerge/blob/performance/test/typescript_test.ts#L355-L361

As a corollary to this, what should happen if i do:

```
text.insertAt(0, 'abc', 'abc', 'abc')
```

The PR does not cover this case and the outcome is

```
Text = ['a', 'b', 'c']
```

where the latter two strings are rejected without an error.

frontend/text.js
lightsprint09: Thanks for the comment. I fixed it and and added a test in dfe1929

lightsprint09:
> As a corollary to this, what should happen if i do:
> 
> ```
> text.insertAt(0, 'abc', 'abc', 'abc')
> ```
> 
> The PR does not cover this case and the outcome is
> 
> ```
> Text = ['a', 'b', 'c']
> ```
> 
> where the latter two strings are rejected without an error.

What do you expect to happen. Should we throw an error or support it it with a result of: `Text = ['a', 'b', 'c', 'a', 'b', 'c', 'a', 'b', 'c']`

nemanja-tosic:
I'm not a maintainer so i wouldn't know.

I'm pointing out how the tests prior to the PR were handling the insertion of strings and it appears to me that the tests weren't asserting the same thing across the board. For example, https://github.com/automerge/automerge/blob/performance/test/typescript_test.ts#L327-L331, is specifying that inserting a string leads to an insertion of a single element, however the assertion doesn't check for that. This PR inserts a string one character at a time. So, it feels that there is a conflict between the previous specs and the behavior this PR introduces, or it's me who is misunderstanding the specs.

Given that. whether or not we throw an error is dependent upon the question "What is the expected behavior for text.insertAt(0, 'abc')?" :)

josharian:
The discussion in #326 seems to have landed at:

> we have a single insertAt method as now, but have it accept an array of characters or a string (which gets split) as second argument

That doesn't seem to be what this PR implements. That plan should also make these calls unambiguous, because the sole arg (it's not variadic any more) is either a string or an array.

lightsprint09:
@josharian 

So the call side would look like the following?

``` js
text.insertAt(0, ["a", "b", "c"])
```
or
``` js
text.insertAt(0, "abc")
```

josharian:
That's how I read it, yes.

lightsprint09:
I close this for now

pvh:
Thanks for a good discussion of the situation and especially @nemanja-tosic and @josharian for review. We've definitely already agreed to do something like this so I hope you'll find a way to clean up this patch so we can merge it, @lightsprint09. If you need help, I have been meaning to work on this so just let me know.

ept:
Thanks for taking the initiative for this change! I've updated the PR to allow to forms of `Text.insertAt()`:

```js
// means Automerge should split the string into individual characters
text.insertAt(index, 'hello')

// takes a string that has already been split;
// Automerge does not do any further splitting.
text.insertAt(index, ['h', 'e', 'l', 'l', 'o'])
```

Passing an array rather than using varargs avoids @josharian's problem of hitting the maximum number of arguments. Also, because the array form does not perform any further splitting, it is suitable for app developers who want to use custom splitting logic (e.g. splitting by Unicode grapheme cluster).

The API also needs to support inserting non-string values (such as marker objects for rich text). For now I've set it up so that the following two are equivalent:

```js
text.insertAt(index, {bold: true})
text.insertAt(index, [{bold: true}])
```

It seems weird that those two calls do the same thing, but I'm not sure what a better alternative would be. One option would be to only allow the latter (where the control object appears inside an array), but that's a bit clumsy to use, and would be inconsistent with the API for inserting items into lists. What do folks think about this? Happy to have a little bit of bikeshedding…

pvh:
I think I lean towards only supporting the latter: it either takes a string or an array of items to insert. All else is incorrect usage. This would prevent problems caused by accidentally passing other unrelated objects in.

ept:
@pvh If we require any non-string values to be wrapped in an array, should we also change the API for lists to be consistent? That API currently works like this:

```js
let x = Automerge.change(Automerge.init(), doc => {
  doc.list = [1, 2, 4]
})
x = Automerge.change(x, doc => {
  doc.list.insertAt(2, 3) // insert the number 3 at index 2
})
// x is now { list: [ 1, 2, 3, 4 ] }
```

To be consistent with the Automerge.Text API we should then also change the list API so that you have to call `doc.list.insertAt(2, [3])` if you want to insert the number 3 into a list — but that seems a bit weird to me. Moreover, it raises the question of what happens when you want to insert a nested list as a list element — would you then have to write `doc.list.insertAt(2, [[...]])`? This doesn't seem very nice. But having the Text and list APIs inconsistent with each other is not nice either.

crn420:
polymorphic-insertAt

crn420:
polymorphic-insertAt

crn420:
@lightsprint09 `

crn420:
#350

crn420:
Duplicate of #

LiraNuna:
What is the expected behavior when splitting emojis? Given the code, I think it will split `👨‍👩‍👦‍👦` into `'👨', ZWJ, '👩', ZWJ, '👦', ZWJ, '👦'`, is this desired?

ept:
Yes, this code will split text into Unicode scalar values, which means some emoji will get broken up. I have previously argued to use grapheme cluster segmentation, which would keep such emoji together as a single string. However, we discussed this recently in automerge/automerge-rs#330 and came to the conclusion that it's okay to split into Unicode scalar values.

PR #363, opened 2021-05-11
Sync fails if one node experiences complete data loss but another node has a stale sync state

dan-weaver:
What I'm trying to simulate here is if one node loses all of it's data including document data as well as any previous sync state information.

The test i've created here goes into an infinite loop.  I believe I'm experiencing this in a real world app but I'm not 100% sure I've captured it accurately in the test.

I'm wondering if this is something application developers should be proactive about (sending a sync message for a blank document) or if this is an actual bug in the sync protocol.

I am able to sync these docs when they both reset their sync state to `initialSyncState()`.  So the question is how should the node that has intact state react to the "bad" sync request from the errant peer?

EDIT: I made the test pass but I'm not sure what kind of damage my "fix" may have caused.  It seems reasonable that if the incoming message is saying "I have nothing" that the entire sync protocol should begin anew?

### DIFF
--- a/backend/sync.js
+++ b/backend/sync.js
@@ -430,7 +430,7 @@ function receiveSyncMessage(backend, oldSyncState, binaryMessage) {
     throw new Error("generateSyncMessage requires a syncState, which can be created with initSyncState()")
   }
 
-  let { sharedHeads, lastSentHeads } = oldSyncState, patch = null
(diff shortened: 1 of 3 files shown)


ept:
Hi @dan-weaver, thank you for this! I've commented on #364. I think your protocol tweak makes sense, so I've merged this PR.

PR #365, opened 2021-05-11
Backend fails to load doc with ambiguous float

vedantroy:
The doc is `{ birds: 3.0 }`. The bytes are generated by creating &
saving the doc in the Rust backend.


orionz:
I added to this and created a new PR.  https://github.com/automerge/automerge/pull/371

PR #366, opened 2021-05-11
Fix wasm tests

jeffa5:
This fixes some of the wasm tests that I'm finding issues with while implementing the new patch format in https://github.com/automerge/automerge-rs/pull/88.

Seems it also brings to light some missing `type: 'value'`s in the js code too. (I'd appreciate someone with more intimate knowledge of the js implementation to help me with the errors here as trying to change a couple of the places in the code makes more things fail).


ept:
This looks great, thank you @jeffa5!

Issue #360, opened 2021-05-08
Big performance regression calling merge() on 1.0.1 preview


josephg:
Sorry the example is a bit long, but this is a simple fuzzer which inserts a bunch of items randomly into 3 simulated peers, and syncs them. In automerge 0.14 on npm this runs reasonably quickly (10000 edits in 5 seconds). Using automerge 1.0.1-preview it runs much more slowly (40 iterations in 5 seconds)

The time is spent in the merge() function - you can see the inflation of merge time as the document grows:

```
merge 0: 2.462ms
merge 1: 2.966ms
merge 2: 5.468ms
merge 3: 5.777ms
merge 5: 7.821ms
...
merge 13: 51.929ms
merge 14: 62.927ms
merge 15: 77.269ms
merge 17: 104.805ms
...
merge 33: 508.224ms
merge 35: 555.231ms
merge 36: 669.424ms
merge 37: 736.736ms
```

The majority of the time (96%) is spent in `decodeColumns`

![image](https://user-images.githubusercontent.com/47413/117555931-166dc980-b053-11eb-9bb5-e2f3155ff1a1.png)

Code:

```javascript
const automerge = require('automerge')
const assert = require('assert')

const amInit = automerge.from({arr: []})

const randInt = (n) => Math.floor(Math.random() * n)

const docs = [
  automerge.merge(automerge.init('01'), amInit),
  automerge.merge(automerge.init('02'), amInit),
  automerge.merge(automerge.init('03'), amInit),
]

let nextItem = 0
const start = Date.now()

let i
for (i = 0;; i++) {
  if (Date.now() - start > 5000) break // Run for 5 seconds
  // console.log(i)
  // if (i % 100 === 0) console.log(i)

  // Generate a random operation
  const d = randInt(docs.length)
  let doc = docs[d]

  const len = doc.arr.length

  // Insert an item
  const content = ++nextItem
  const pos = randInt(len + 1)
  doc = automerge.change(doc, d => {
    d.arr.splice(pos, 0, content)
  })

  docs[d] = doc

  // Pick a pair of documents and merge them
  const a = randInt(docs.length)
  const b = randInt(docs.length)
  if (a !== b) {
    console.time(`merge ${i}`)
    docs[a] = automerge.merge(docs[a], docs[b])
    docs[b] = automerge.merge(docs[b], docs[a])
    console.timeEnd(`merge ${i}`)
    assert.deepStrictEqual(docs[a], docs[b])
  }
}

[...]

PR #368, opened 2021-05-13
Improve performance of repeated calls to `Automerge.merge()`

ept:
Previously, `Automerge.merge(localDoc, remoteDoc)` would get all changes from `remoteDoc` and apply them to `localDoc`, relying on the built-in change deduplication to ignore any changes already present in `localDoc`. This works but is quite slow. This PR changes `merge` to get only those changes in `remoteDoc` that are not already present in `localDoc`, and applies those.

On @josephg's benchmark in #360 this improves performance by a factor of 15 on my machine:

* Automerge 0.14.1: about 7,100 iterations in 5 seconds
* Automerge 1.0.1-preview.1: about 300 iterations in 5 seconds
* With this PR: about 4,700 iterations in 5 seconds

Note this PR introduces a new backend function `getChangesAdded()`, which will also need to be implemented in the Rust backend.

### DIFF
--- a/@types/automerge/index.d.ts
+++ b/@types/automerge/index.d.ts
@@ -145,6 +145,7 @@ declare module 'automerge' {
     function getAllChanges(state: BackendState): BinaryChange[]
     function getChangeByHash(state: BackendState, hash: Hash): BinaryChange
     function getChanges(state: BackendState, haveDeps: Hash[]): BinaryChange[]
+    function getChangesAdded(state1: BackendState, state2: BackendState): BinaryChange[]
     function getHeads(state: BackendState): Hash[]
     function getMissingDeps(state: BackendState, heads?: Hash[]): Hash[]
     function getPatch(state: BackendState): Patch
--- a/backend/backend.js
+++ b/backend/backend.js
@@ -269,6 +269,20 @@ function getChanges(backend, haveDeps) {
 }
 
 /**
+ * Returns all changes that are present in `backend2` but not in `backend1`.
+ * Intended for use in situations where the two backends are for different actors.
+ * To get the changes added between an older and a newer document state of the same
+ * actor, use `getChanges()` instead. `getChangesAdded()` throws an exception if
+ * one of the backend states is frozen (i.e. if it is not the latest state of that
+ * backend instance; this distinction matters when the backend is mutable).
+ */
+function getChangesAdded(backend1, backend2) {
+  const opSet1 = backendState(backend1).get('opSet')
+  const opSet2 = backendState(backend2).get('opSet')
+  return OpSet.getChangesAdded(opSet1, opSet2)
+}
+
+/**
  * If the backend has applied a change with the given `hash` (given as a
  * hexadecimal string), returns that change (as a byte array). Returns undefined
  * if no change with that hash has been applied. A change with missing
@@ -304,5 +318,5 @@ function getMissingDeps(backend, heads = []) {
 
 module.exports = {
   init, clone, free, applyChanges, applyLocalChange, save, load, loadChanges, getPatch,
-  getHeads, getAllChanges, getChanges, getChangeByHash, getMissingDeps
+  getHeads, getAllChanges, getChanges, getChangesAdded, getChangeByHash, getMissingDeps
 }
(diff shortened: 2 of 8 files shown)


josephg:
Thanks for working on this! Its weird - I'm seeing very different performance numbers from you:

* 1.0.1-preview-1: In 5 seconds ran 36 iterations
* 0.14.2: In 5 seconds ran 10228 iterations
* This faster-merge branch: In 5 seconds ran 331 iterations

So this helps, but there's still more than an order of magnitude slowdown compared to 0.14.2.

```
$ node --version
v16.1.0   # (on linux, if that matters somehow)
```

```
$ grep version node_modules/automerge/package.json 
  "version": "1.0.1-preview.1",
$ node am.js
In 5 seconds ran 36 iterations
```

```
$ yarn add --dev automerge@0.14.2
...
Done in 0.11s.
$ grep version node_modules/automerge/package.json
  "version": "0.14.2",
$ node am.js                                
In 5 seconds ran 10228 iterations
```

```
[leaf] ~/3rdparty/automerge:faster-merge ✓ $ yarn build
...
Done in 2.92s.
[leaf] ~/3rdparty/automerge:faster-merge ✓ $ git log -n1
commit 4e2f32e832b0a21310989e13901a7cb76c085b8e (HEAD -> faster-merge, origin/faster-merge)
[leaf] ~/3rdparty/automerge:faster-merge ✓ $ node am.js 
In 5 seconds ran 331 iterations
```

josephg:
Aaah for some reason the performance difference comes from webpack. If I make this change to package.json:

```
  "main": "src/automerge.js",
```

Then I see performance numbers similar to you!

```
[leaf] ~/3rdparty/automerge:faster-merge ✗ $ node am.js      
In 5 seconds ran 7374 iterations
```

josephg:
Delving further, the performance returns if I remove babel from the webpack.config.js. It looks like one of the JS transforms / polyfills that babel is doing axes performance.

ept:
Oh wow. Good find. When running with the webpack output (`dist/automerge.js` after running `yarn build`) it does indeed get massively slower. Flamegraph shows the code is spending almost all of its time in [`Decoder.readInt64()`](https://github.com/automerge/automerge/blob/9943f4fa5725a9ad7c21ead141fb703615e2b0cd/backend/encoding.js#L458-L496). I manually compared the webpack output for that function to the original code and did not find any significant differences. I guess that maybe `Uint8Array` is getting replaced by a slow polyfill?

How do you change `webpack.config.js` to remove Babel? I don't know how this stuff works…

skokenes:
Seems like its coming from core-js https://github.com/zloirock/core-js#caveats-when-using-typed-arrays-polyfills

Definitely interested in this as our UI package that uses Automerge does go through its own bundling process with babel/core-js. Would love to know how we do that without accidentally hitting this typed array polyfill issue

josephg:
The babel configuration is these lines in webpack.config.js:

```
  module: {
    rules: [
      {
        test: /\.js$/,
        exclude: /node_modules/,
        use: {
          loader: 'babel-loader',
          options: {
            presets: [
              ['@babel/preset-env', { targets: 'defaults' }],
            ]
          }
        }
      }
    ]
  }
```

This entry in the "rules" array is naming babel-loader, and telling it to transpile the javascript to work on a list of targets that [browserlist](https://github.com/browserslist/browserslist#queries) considers reasonable "defaults". And it looks like that includes some super old web browsers which don't support typed arrays.

Personally I think its weird for an npm module to do this work. If people want to use automerge in their web app, they should just compile / transpile it themselves to support whatever list of web browsers they decide to support. A lot of work has gone into making this sort of thing work reasonably well through npm anyway.

Anyway, options, in decreasing severity:

1. Remove webpack entirely. Just have package.json reference `src/automerge.js`. Automerge doesn't need a build step. Users who want compiled javascript can make their own.
2. Remove babel from webpack. Just make the webpack config have an empty `rules: []` array.
3. Change the babel preset target from `"defaults"` to either `"> 1%, not dead"` or `"maintained node versions"` (which is my preference). The former says "allow features only if they're supported in every browser with more than 1% usage share". It looks like thats enough for typed arrays not to be polyfilled. The latter says "make sure it'll work on the oldest version of nodejs in the LTS window". (Which at the moment iirc is node 10+, maybe node 12+). Given we're publishing to npm, that makes lots of sense to me. Either of those settings seems to do pretty much the same thing - webpack will stop putting in a typed array polyfill and performance gets way better.

skokenes:
I agree with @josephg , I think AM could at least get away with dropping the `@babel/preset-env` preset all together as it should be the consumer's job to bundle for their target environment

josephg:
Yeah I agree. But just to be clear, removing the `@babel/preset-env` line only makes babel revert to its default transpile target. Which still unfortunately polyfills typed arrays.

If we want to leave JS target selection up to our users, we should either remove babel entirely (`module: { rules: [] }`) or remove the webpack build step entirely, and as I said above just reference `src/automerge.js` directly from package.json. The package works fine as it is.

ept:
I think it's useful to keep the webpack config and the build step as it produces a self-contained JS file that can be used directly in browsers without further work. It's also used by the browser-based tests that are run as part of Automerge's CI. However, I'm happy for that JS file to only work in modern browsers, and if people want to target older browsers I think it would be fair to require them to perform their own transpiling. Thus, I think we can remove Babel (which also has the nice property of making the `node_modules` directory a bit smaller). To me it also seems fine to point package.json directly at the original source (I never quite understood why it should point at the webpack output). I have added commits to that effect to this PR.

If anybody disagrees on how this should be done, please speak up! I am totally clueless about JS packaging…

PR #371, opened 2021-05-14
Strict Number  Types 

orionz:
Superseding PR https://github.com/automerge/automerge/pull/365

This builds on the failing test Ved created by adding 4 strict number datatypes.  "int", "uint", "float32" and "float64".  I try to make the datatype be optional but always included when a change is decoded.  This made for a fairly small patch.  I worry that once implemented in rust a bunch of tests will fail because decodes will happen in that code path that do not happen for the js side, specifically the datatypes that emerge in patches.  We may need to make it non-optional for numbers.  The other two shortcomings I see are (1) the frontend in javascript has no way to specify a specific number type.  I can't set `doc.value = 1` to a float.  Also some tests to exercise all the different number types are needed.  This format change will require all other implementations (swift etc) to update accordingly.


orionz:
Ok.  I added some tests but curiously I could find no float value would meet the criteria we currently have for a 32bit float.  We could drop support for float32 and make all floats 64bit - due note tho that outside of javascript f32 could be useful.

ept:
I think it would be a good idea to keep support for float32. I think `0.5` meets the criteria for a 32-bit float.

Can you let me know when this PR is ready to review?

orionz:
Ok.  I created classes to let people make Float32 etc the same way Timestamps and Coutners are made.  All tests passing.  Rust has a matching PR.  I'd like to address the multi-insert with different types in a different PR.

test/backend_test.js
```diff
@@ -569,11 +569,21 @@ describe('Automerge.Backend', () => {
       })
     })
 
+    it('should thorw an error if the datatype does not match the values', () => {
```
vedantroy, line 608: nit: s/thorw/throw

ept:
Sorry that it's taken me so long to review this, @orionz @vedantroy. This looks great! I've made some very minor tweaks and merged it.

PR #377, opened 2021-05-16
guard against null 'theirHeads'

dan-weaver:
In my experimentation with they sync protocol I've come across what may be another edge case.  `theirHeads` is sometimes null here when there are pending changes in the doc.  Not sure if my implementation is bad.  Should theirHeads just default to `[]`?


ept:
Hi @dan-weaver, thanks for the bug report. `theirHeads` being null is a valid state, so we need to fix this. However, I think the logic should be different. Could you please try 77bd0e85f7f47cde85d55c52a2c40a3e16d12509 and see if it fixes the issue for you?

dan-weaver:
@ept yep works for me thanks!

PR #382, opened 2021-05-18
Failing relative ops in the same change test

jeffa5:
I'm not completely sure that this should be failing but it seems like it should be.

The issue is that the last action of `change2` is trying to insert after the update that has just occurred. I would assume that the prior update also updates the `elemId` of that index and so we should be able to reference it.

### DIFF
--- a/test/backend_test.js
+++ b/test/backend_test.js
@@ -760,6 +760,22 @@ describe('Automerge.Backend', () => {
         }}}}
       })
     })
+
+    it('should allow list operations relative to ops in the same change', () => {
+      const actor = uuid()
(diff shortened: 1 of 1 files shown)


jeffa5:
Changing the last `elemId` to `2@${actor}` passes.

jeffa5:
Or I suppose that list element should still have the `elemId: 2@${actor}` if they are supposed to be immutable?

jeffa5:
I think I've found the problem in the Rust frontend. Sequences elements should be linked to the operation that created them and updates don't change those.

ept:
I think it's correct for the last op to have an elemId of `2@${actor}`. Op `3@${actor}` is not an insertion op so it cannot be used as the elemId of another insertion op.

PR #385, opened 2021-05-20
theirHeads should be empty array rather than null

okdistribute:
Ran into this bug while restarting from a persisted state without saving the sync state locally between syncronizations.


okdistribute:
Realized this was caught two days ago by martin! 🤓

PR #386, opened 2021-05-20
Increase threshold in deflate test

jeffa5:
This is giving occasional errors in the rust/wasm tests as it seems to
sometimes be ~101/102.


jeffa5:
Since there seems to be quite a large difference between the js and rust output here I'm thinking we have ourselves a bug...

ept:
It's possible they are just using different compression levels or algorithms for deflate. I reckon we could safely bump the number up to 200 if that makes the test more robust.

Issue #370, opened 2021-05-14
Publishing the library to npm


scotttrinh:
We've had a long conversation in Slack to discuss the best way to publish the library to npm, and I wanted to provide that context in a more visible public forum.

## The problem

Automerge can be run in the browser and also in Node, however, it does use some language features that are not supported by _all_ browsers and node versions. The current situation imposes an implicit minimum required environment by using language features without transpiling. This can lead to some confusing issues and "bugs" by downstream consumers.

There is also an issue with polyfills degrading performance (see #368 and #360 for instance), we need to pick a target environment that doesn't penalize modern targets that support the features used here.

## Solutions

The current state of publishing packages to npm has evolved a lot lately. npm packaging support is more sophisticated, node now supports ES modules in addition to CommonJS, and bundling for web applications is more controversial with HTTP/2 becoming ubiquitous.

The two parts of a solution here involve: transpiling the code as-written to a reasonable (and explicit) target environment, and exporting useful module paths (and bundles?) for different environments.

### Proposal

I propose that we publish the following artifacts:

1. A transpiled, unbundled ES module path
2. A transpiled, unbundled CommonJS module path
3. (Maybe?) A transpiled, bundled UMD module

That should cover the following environments:

1. Modern browser consumers can choose to use ES modules directly without a bundling step, or use the UMD bundle if they prefer.
2. Consumers that wish to support older browsers can use the prebundled UMD.
3. Node can use either ES modules or CommonJS, depending on the consumers preferences and tooling/support.

### Remaining question: What target environment to specify?

We will likely want to keep the existing tooling around transpilation and bundling, so that leads us toward using babel with the [`babel-preset-env`](https://babeljs.io/docs/en/babel-preset-env) plugin for specifying the transpilation target, and using Webpack to bundle the UMD module.

In order to use the `babel-preset-env` plugin, we need to specify our targets, and I believe the consensus on a requirement so far is to specify a target that will do the minimum amount of poly-filling and transformation as possible but allows for a nice development experience still. Let's discuss!

skokenes:
Funny enough, this is already biting me today in a UI project because automerge uses node's builtin `string_decoder`, which is not available in browsers

scotttrinh:
@skokenes 

Do you think either the UMD module or the ESM module would solve your issue if we set up the transpilation based on the targets I mentioned above?

skokenes:
Probably?

[...]

PR #396, opened 2021-05-26
Add `eslint-plugin-compat` and update linting config

scotttrinh:
Adds linting to detect usage of features that are not natively supported by our target platforms. Sets those platforms using https://github.com/browserslist/browserslist but we should negotiate what those platform ought to be before merging this—I picked this list as a sensible starting point.

We should also update the docs to let consumers know what sort of environments are supported.

**Note**: @babel/preset-env also uses this same setup as the target of its transpilation and polyfilling, so we will want to check the output of the bundling and transpilation to ensure it does not include any performance regressions. See #368 for some babel-related discussion that is relevant here.

Closes https://github.com/automerge/automerge/issues/370

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -2,14 +2,14 @@
     "env": {
         "browser": true,
         "commonjs": true,
-        "es2021": true,
+        "es2015": true,
         "node": true,
         "mocha": true
     },
     "ignorePatterns": "dist/**",
-    "extends": "eslint:recommended",
+    "extends": ["eslint:recommended", "plugin:compat/recommended"],
     "parserOptions": {
-        "ecmaVersion": 12
+        "ecmaVersion": 2015
     },
     "rules": {
         "accessor-pairs": "error",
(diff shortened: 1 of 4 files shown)


.eslintrc.json
```diff
@@ -2,14 +2,14 @@
     "env": {
         "browser": true,
         "commonjs": true,
-        "es2021": true,
+        "es2015": true,
```
scotttrinh, line 5: Cranking down our ES version to catch any advanced feature usage at the parser-level. We can turn this up if we end up using features defined outside of `es2015`.

package.json
```diff
@@ -55,5 +56,10 @@
   },
   "resolutions": {
     "karma-sauce-launcher/selenium-webdriver": "4.0.0-alpha.7"
-  }
+  },
+  "browserslist": [
+    "defaults",
+    "not IE 11",
+    "maintained node versions"
+  ]
```
scotttrinh, line 64: Totally open for suggestion here.

ept:
Hi @scotttrinh, thank you! This config looks reasonable to me.

At the moment when running webpack I get `Error: Universal Chunk Loading is not implemented yet`. Is that just me?

ept:
Okay, this seems fine to me — thank you for your help @scotttrinh!

PR #397, opened 2021-05-27
Do not chunk the output in webpack config

scotttrinh:
The browser tests broke during the refactor at #396 complaining about not being able to iterate over undefined "chunks":

```
  Uncaught TypeError: Cannot read property 'webpackChunkAutomerge' of undefined
```

That reminded me of the issue that was noted in that original PR where setting this `chunkLoading` property didn't seem to make a difference for running webpack, but it looks like it _does_ matter for actually using the bundle! 😅 

I've run `yarn run browsertest` locally and it seems to work, so hopefully this fixes the issue.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -10,6 +10,8 @@ module.exports = {
     path: path.resolve(__dirname, 'dist'),
     // https://github.com/webpack/webpack/issues/6525
     globalObject: 'this',
+    // https://github.com/webpack/webpack/issues/11660
+    chunkLoading: false,
   },
   devtool: 'source-map',
   module: {rules: []},


ept:
Thanks, I was about to ask you for help! This also fixes the browser tests for me, but I noticed that after this PR Karma is only running 60 tests in each browser, whereas previously it was running 562. I think it's only running the tests in `test/typescript_test.ts` but not those in `test/*test.js`. Do you know how to get it to run all the tests again?

scotttrinh:
Oh interesting. Sure, lemme take a look and see if I can figure out what's happening.

ept:
I will go ahead and merge this in order to make Travis CI happy again, so that I can make a release, but let's try to fix the test runner in a follow-up PR.

PR #412, opened 2021-06-17
Require actor ids to be uuid length

jeffa5:
The actor id is required to be a uuid but the length is not currently enforced.

I'm not aware of a documented reason to not observe a consistent length but do let me know if it was intentional.

### DIFF
--- a/frontend/index.js
+++ b/frontend/index.js
@@ -20,8 +20,8 @@ function checkActorId(actorId) {
   if (!/^[0-9a-f]+$/.test(actorId)) {
     throw new RangeError('actorId must consist only of lowercase hex digits')
   }
-  if (actorId.length % 2 !== 0) {
-    throw new RangeError('actorId must consist of an even number of digits')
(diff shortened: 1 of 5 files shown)


jeffa5:
Ah, just seen https://github.com/automerge/automerge-rs/pull/178#issuecomment-863544241

lightsprint09:
As an user of the API I don't understand why an Actor needs to be fixed size. Maybe I use some Identity Managment System which provides me stable IDs which I can use as ActorId but are not UUID sized.

jeffa5:
Well it is documented as a UUID so I anticipated that it should be one but turns out that is a bit misleading

MeneDev:
I'd prefer a change to the documentation rather than the code. As @ept argued in https://github.com/automerge/automerge-rs/pull/178 this could be unnecessarily limiting.
Personally I thought about using git commit hashes (40 hex digits) as actors.

PR #420, opened 2021-08-10
doc: use recent JS idiom

efx:
I find using recent JavaScript idioms in README.md documentation
increases the appeal for many JavaScript developers.


ept:
Thanks @efx!

PR #426, opened 2021-08-16
Include startOp into TypeScript definition of Change

begor:
(follow-up on #424)

When generation `automerge.Change`, we need to include `startOp` field to be able to encode it using `automerge.encodeChange`. In order to do this, we need to include `startOp` into TypeScript interface of `Change`.


ept:
Thanks @begor!

Issue #429, opened 2021-08-20
Generating `deps` of `Change`


begor:
Hello!

From documentation it's a bit unclear how we can generate `deps` field of a `Change` object:
```
The responsibilities between frontend and backend are assigned as follows:

* The frontend assigns sequence numbers and operation IDs to locally generated
  changes, and produces a JSON representation of the change.
* The backend translates the JSON change into the binary encoding, **computes its
  hash, and fills in the dependency hashes as described below.**
```

But it doesn't look like this for me, reading existing frontend code, in `frontend/index.js:makeChange()` we [actually set](https://github.com/automerge/automerge/blob/main/frontend/index.js#L90) deps. 

Moreover, when I'm writing my own frontend for automerge, it looks like I should fill up `deps` on my own, because backend actually checks my `startOp` against them [here](https://github.com/automerge/automerge/blob/main/backend/op_set.js#L401). 

So there are two questions, basically:

1) Who should fill up `deps` for a `Change`? Documentation states that this is a backend responsibility, but code looks different to me. 
2) What the exact semantics of `deps`? Suppose I generate a change, what `deps` should it have? Only the "head" of history (as with `automerge.Backend.getHeads`)? Or hash of every change that our new change "references" (e.g. it mutates the object that was created by some other change)?

ept:
We went back and forth on this a bit, so the documentation is not quite up-to-date with the code. In the current implementation, the frontend fills in deps for any changes received from other actors, but the [backend adds a dependency](https://github.com/automerge/automerge/blob/4e2f32e832b0a21310989e13901a7cb76c085b8e/backend/backend.js#L128-L149) on the previous change by the local actor. See the comment in the linked code for an explanation.

Deps only needs to include the heads (which is typically just a single hash, except in the case of a "merge commit" which would have two or more deps). All changes that can be transitively reached through the chain of deps are implicitly included as dependencies, so there is no need to reference specific changes in the past.

begor:
> Deps only needs to include the heads (which is typically just a single hash, except in the case of a "merge commit" which would have two or more deps). All changes that can be transitively reached through the chain of deps are implicitly included as dependencies, so there is no need to reference specific changes in the past.

Currently I'm doing something like this for every change:
```
const change = {
     ...
     deps: automerge.Backend.getHeads(backendState)
```

The change then goes straight to `automerge.Backend.applyChanges(...)`. 
Is this a correct approach?

ept:
Yes I think that should be fine.

[...]

PR #430, opened 2021-08-20
Include `hash` to TypeScript definition of `Change`

begor:
(follow-up on #424 and #426, sorry for that excessive granularity, I'm discovering missing fields as I go deeper into B-F protocol)

Related to #429.

`hash` field stores SHA-256 of a binary encoding of change. It's used to fill up `deps` of the following changes. However, it's missing from TS definition of `Change`, but present in the plain JS object:

```
{
        "actor": "76606470dee9469391163cb3c43c220a",
        "seq": 2,
        "startOp": 2,
        "time": 1629452632,
        "message": "",
        "deps": [
            "f81149f2a77edb5425d64966151d3f8d66a30a907ed29760a47e4c2805439169"
        ],
        "hash": "ae910c6f52b09c9f256be9cc9ba4f5925c8fcb5066c81a4faac226e547e4a8df",
        "ops": [...]
```

So here we expose it, but make it optional, since it's backend responsibility to fill it.

### DIFF
--- a/@types/automerge/index.d.ts
+++ b/@types/automerge/index.d.ts
@@ -220,6 +220,7 @@ declare module 'automerge' {
     time: number
     seq: number
     startOp: number
+    hash: Hash
     deps: Hash[]
     ops: Op[]
   }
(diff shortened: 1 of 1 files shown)


ept:
Yeah this makes sense. In principle every change has a hash, but it might not yet have been computed, so it's best to make it optional.

PR #433, opened 2021-09-22
Eliminate loop over ops in Context.nextOpId()

luchak:
We have a use case where we need to copy large documents, containing
as many as ~100k objects nested in various ways. This was quite
slow, and profiling revealed that we were spending a lot of time in
Context.nextOpId(). This change tracks nextOpNum as a property on the
context, and updates it in addOp() when new ops are pushed.

The change here drastically speeds things up for us: after this, copying
is still a little slow, but the situation is much improved, and the
bottleneck has moved from generating the op list to applying it.

The new strategy should be functionally equivalent to the old one, and
this change passes tests (at least when I run them), but I very well may
have missed something, so feedback is welcome!

### DIFF
--- a/frontend/context.js
+++ b/frontend/context.js
@@ -16,7 +16,7 @@ const uuid = require('../src/uuid')
 class Context {
   constructor (doc, actorId, applyPatch) {
     this.actorId = actorId
-    this.maxOp = doc[STATE].maxOp
+    this.nextOpNum = doc[STATE].maxOp + 1
     this.cache = doc[CACHE]
     this.updated = {}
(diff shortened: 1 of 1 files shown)


luchak:
Looks like there was a timeout in connecting to the browser test container in CI. I suspect this is a transient issue with CI, though I'm not sure I have the ability to re-run the step myself.

ept:
Hi @luchak, thank you for this nice improvement! The CI failure is unrelated to your change (I recently moved from Travis CI to GitHub Actions and haven't quite ironed out all the issues yet).

Issue #257, opened 2020-05-13
 Proxy polyfill does not support trap 'deleteProperty' 


leehambley:
Note: Opening this issue here just as a heads-up. 

I don't know that I expect it to be fixed, but I did want to have raised it.

I have to support Mobile Safari 9 (internal version `601.2` in the UA string) and it completely lacks support for the `Proxy` object. No big deal, we can polyfill with `proxy-polyfill`.

However, `proxy-polyfill` doesn't support a specific feature that is used (and none of other pollyfils I could find do, either)

```
TypeError: Proxy polyfill does not support trap 'deleteProperty'
  at trap(../node_modules/proxy-polyfill/src/proxy.js:60:19)
  at mapProxy(../webpack:/Automerge/frontend/proxies.js:195:14)
  at rootObjectProxy(../webpack:/Automerge/frontend/proxies.js:220:10)
  at Frontend(../webpack:/Automerge/frontend/index.js:274:12)
  at change(../webpack:/Automerge/src/automerge.js:39:37)
  at quizReducer(quizReducer.ts:234:20)
  at dispatch(../node_modules/redux/es/redux.js:213:22)
  at apply(../node_modules/redux/es/redux.js:299:3)
  at enhancer(../node_modules/redux/es/redux.js:629:31)
  at createStore(../node_modules/redux/es/redux.js:79:12)
  at call(quizStore.ts:72:17)
  at tryCatch(../node_modules/regenerator-runtime/runtime.js:45:40)
  at _invoke(../node_modules/regenerator-runtime/runtime.js:271:22)
  at call(../node_modules/regenerator-runtime/runtime.js:97:21)
  at tryCatch(../node_modules/regenerator-runtime/runtime.js:45:40)
  at invoke(../node_modules/regenerator-runtime/runtime.js:135:20)
  at handler(../node_modules/regenerator-runtime/runtime.js:145:13)
  at fn(../node_modules/core-js/modules/es.promise.js:114:22)
  at r(../node_modules/core-js/internals/microtask.js:26:9)
  at promiseReactionJob([native code])
```

Safari 0.1 versions newer adds support for this, so I can definitely work around this, and I'm starting to discover a base-line for what I can, or cannot support. 

Feel free to close if this is/was a known issue and there's no easy workaround. I don't know enough about `Proxy` or the polyfills to know if ther is an alternative to the delete trap.

---

- https://caniuse.com/#feat=proxy
- https://github.com/GoogleChrome/proxy-polyfill

[...]

PR #438, opened 2021-10-03
[RFC] Add alternative frontend API without proxy

zhang123cnn:
* This is only a RFC PR for comments. Please do not merge. *

Hey folks,

My name is Xiaomeng and I am a software from Facebook. My colleague @camiboj and I are working on an experimental project to leverage automerge to decentralise state management in our infrastructure.

Automerge uses JS Proxy extensively for its front end API. However, in FB, we need to support multiple JS runtime which does not support Proxy.

I can see a similar question has been raised here before.

https://github.com/automerge/automerge/issues/257

@ept advised us we could implement an alternative API which does not use proxies. So we went ahead and implemented a minimal version for ourselves.

It is very similar to current API. Because we could not use proxy, the new API requires lib users to be explicit on built-in functions for list, map. For example, You cannot do `doc['key1'] = 'value1'`. Instead you need to do `doc.set('key1', 'value1')`. Check our unit test file test/proxies_polyfill_test.js for more examples in code.

Our fork worked for us, but ideally we would prefer not to keep a permanent fork. So we have been wondering if it is possible to merge this alternative API back to automerge repo. It would help us not to maintain a fork and maybe help others potential automerge users who also lives with similar constraints. 

What do you think?

CC @ jtannady

### DIFF
--- a/frontend/index.js
+++ b/frontend/index.js
@@ -2,13 +2,14 @@ const { OPTIONS, CACHE, STATE, OBJECT_ID, CONFLICTS, CHANGE, ELEM_IDS } = requir
 const { isObject, copyObject } = require('../src/common')
 const uuid = require('../src/uuid')
 const { interpretPatch, cloneRootObject } = require('./apply_patch')
-const { rootObjectProxy } = require('./proxies')
+const { rootObjectProxy, setFacebookSyntax } = require('./proxies')
 const { Context } = require('./context')
 const { Text } = require('./text')
 const { Table } = require('./table')
 const { Counter } = require('./counter')
 const { Float64, Int, Uint } = require('./numbers')
 const { Observable } = require('./observable')
+const { ProxyPolyfill } = require('./proxy_polyfill')
 
 /**
  * Actor IDs must consist only of hexadecimal digits so that they can be encoded
@@ -158,6 +159,13 @@ function applyPatchToDoc(doc, patch, state, fromBackend) {
     state.maxOp = Math.max(state.maxOp, patch.maxOp)
   }
   return updateRootObject(doc, updated, state)
+}
+
+/**
+ * This function will set syntax defined by `ListProxyPolyfill`/`MapProxyPolyfill` as frontend interface
+ */
+function facebookSyntaxOn() {
+  setFacebookSyntax(true)
 }
 
 /**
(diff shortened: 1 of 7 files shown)


ept:
Hi @zhang123cnn! Great to see you've gone ahead with this. I understand the need for a proxy-free API, and I'd be happy to have such an API be part of what we maintain in the Automerge project and its official releases.

We'd probably want to iterate a bit on the details of the API before committing to it. From a code structure point of view, it might make sense to separate the proxy-based frontend and the proxy-free frontend into two separate directories, and allow users to import the one they want, but I'd want to get some more opinions from people with more experience in JS packaging.

Another thing I'm thinking about is that @orionz is currently working on a redesign of the frontend-backend API with a view towards improving performance and making it easier to implement frontends. Assuming that works out, a proxy-free API will probably become a fairly small amount of code, since most of the current frontend complexity will move to the backend.

I'd also love to hear more about what you're doing with Automerge at Facebook!

zhang123cnn:
@ept Thanks for your reply!

> I'd be happy to have such an API be part of what we maintain in the Automerge project and its official releases.

This is great news!

> it might make sense to separate the proxy-based frontend and the proxy-free frontend into two separate directories

This makes sense to me. The majority of the code is already in a standalone js file frontend/proxy_polyfill.js. We could certainly put them into a different directory and separate the entry point from the original proxy.js file.

We currently have a global function `setFacebookSyntax` to switch proxy-based API to non-proxy-based one. I guess we would like to rename it to something more general.

>  Assuming that works out, a proxy-free API will probably become a fairly small amount of code

This is also great! Do you know what would be the rough timeline on this? We do want to ship things with automerge in a few weeks so we would prefer to adapt our code to new design slightly later. 

> I'd also love to hear more about what you're doing with Automerge at Facebook!

In facebook, we are working on a new state management API for AR effects running in video calls. We use automerge because we want all state sync logic happening on client side to better protect user's privacy. We have some interesting use case. Once our announcement is out, we are probably come back to automerge community with more questions and thoughts.

About next steps, I guess we could just create a formal PR to start the iteration. CC @camiboj We do hope to release our API soonish. So please let us know about anything we need to amend to proceed. 

Thanks in advance.

ept:
Great. Please do share your announcement when it's out!

As a first step, before we get into the code, can I suggest that you write up a document explaining the API you're proposing? I know you have the tests, but I think a document would help facilitate a community discussion about the API design. It would be good to include a discussion of the edge cases that arise, such as: how do you deal with nested objects? conflicts? how do you handle the various datatypes that Automerge supports, such as lists, text, and counters?

I'm not sure when the frontend-backend API changes will happen — it's a fairly complex project, since it affects most of the codebase, and requires coordination between the JS and Rust implementations of Automerge. My guess is that it's a couple of weeks away.

PR #439, opened 2021-10-08
Proxy Free API

camiboj:
This is the PR created to start the iteration for including a new API without a proxy as discussed on #438 (cc @zhang123cnn @ept). *

# Proxy Free API 
Automerge uses JS Proxy extensively for its front-end API. However, to be able to support multiple JS runtime which does not support `Proxy` you can use the **Proxy Free API**.

This API does not modify the way automerge handles conflict or nested objects. It is only modifies the `Object` and `Array` APIs.

You can find documentation on this new API on `proxy_free.md`

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -60,7 +60,7 @@ describe('Automerge', () => {
     })
   })
 
-  describe('initialization with facebook syntax', () => {
+  describe('initialization with Proxy Free API', () => {
     after(() => {
       setProxyFree(false)
     })
--- a/test/frontend_test.js
+++ b/test/frontend_test.js
@@ -41,7 +41,7 @@ describe('Automerge.Frontend', () => {
     })
   })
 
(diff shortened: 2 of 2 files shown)


test/test.js
```diff
@@ -59,6 +60,28 @@ describe('Automerge', () => {
     })
   })
 
+  describe('initialization with facebook syntax', () => {
```
zhang123cnn, line 63: could we remove facebook here for the test description?

zhang123cnn:
Hey @ept thanks for your comment. I think that makes sense to us. We can wait until the redesign is finished to review this PR.

At the same time, do you mind if we create a public fork in our facebook github organiztion? We need to host this forked code in github for our upcoming release. Having a public fork would unlock us right now. We will delete that fork once this PR is merged back to original repo though.

Please let us know what you think. Many thanks

CC @jtannady

Issue #374, opened 2021-05-14
API for initializing a document with a 'null' author and timestamp? 


HerbCaudill:
Suppose Alice initializes a document as follows: 

 ```ts
let doc = A.from({ todos: [] })
```

Alice adds some todos: 

```ts
doc = A.change(doc, s => s.todos.push('Learn Automerge'))
doc = A.change(doc, s => s.todos.push('Profit'))
```

Now Bob initializes his document using the same code as Alice: 

 ```ts
let doc = A.from({ todos: [] })
```

Alice and Bob connect using the new sync protocol, perhaps using an implementation like this one. https://github.com/HerbCaudill/todomvc-automerge/blob/master/src/AutomergeSync.tsx

Half of the time, they will converge on Alice's list with two items; the other half, they will converge on Bob's empty list.

I asked about this on Slack and @ept suggested initializing the document with a "null" authorId like `0000`, and a timestamp of `0`. 

<details>
  <summary>conversation from slack fwiw</summary>

> @herbcaudill :
> Continuing my exploration of Automerge 1, I've made a very minimal TodoMVC app in React, which can sync instances in multiple browsers using @localfirst/relay for networking. Code is here https://github.com/HerbCaudill/todomvc-automerge
>
> In my testing, if you make a bunch of changes to the list in one browser and then open another browser window, there's some % chance that the first browser's state will be overwritten with the default state from the second browser.
>
> IIRC, the way I'm doing this wouldn't have worked in Automerge 0.x : each client starts out with this as their state -
>
> ```ts
> const defaultState = A.from<State>({
>   todos: [
>     { id: '0', completed: true, value: 'Learn React' },
>     { id: '1', completed: false, value: 'Learn Automerge' },
>     { id: '2', completed: false, value: 'Profit' },
>   ],
> })
> ```
>
> And the sync protocol lets them reconcile their documents, which in theory started out divergent and shouldn't be able to converge.
>
> I guess I need to have some kind of flag to indicate that I'm still at the default state, and if that's the case then defer to a peer who has already made edits? I've solved this in the past with Automerge 0.x but I'm not sure if I'm thinking about it in the right way & my brain is tired
>
> @ept :
> Rather than using `A.from`, maybe a good option would be to initialise all clients with the same initial change. If the actorId and the change timestamp are hardcoded, the rest of the change generation should be deterministic, and so all clients should have an identical initial change hash.
>
> @herbcaudill :
> hm ok that makes sense. how do I hardcode the timestamp?
>
[...]

PR #445, opened 2021-11-10
Failed test with initial schema hack

okdistribute:
This test fails after I save and then re-load the document. If you remove lines 1445 and 1446, the tests pass. If you port this test to preview.4, the test passes. So this seems to be a bug somewhere between preview.4 and preview.5. I did a git bisect and found that hash 325f9e8f7d4a16ed7a168494161cecf47185aea9 is the first bad commit.

This is a hack for supporting branching and merging from unrelated documents that were created within the same app. It is not directly supported by the Automerge API, which is probably how the bug snuck through. I think it's important to make this hack more automatic in the Automerge API. If I get some time to think about it I can say more or make a suggestion about how the API should work. Without this, there is no way to get reliable behavior for merging two branched documents that don't start with a single author.

```
√ automerge % npm test test/test.js

  Automerge
    changes API
      1) should apply changes from another document initialized with the same schema


  0 passing (37ms)
  1 failing

  1) Automerge
       changes API
         should apply changes from another document initialized with the same schema:
     RangeError: Expected seq 2, got seq 1 from actor 0000
      at applyChanges (backend/new.js:1536:15)
      at BackendDoc.applyChanges (backend/new.js:1778:35)
      at Object.applyChanges (backend/backend.js:29:23)
      at Object.applyChanges (src/automerge.js:93:37)
      at Context.<anonymous> (test/test.js:1448:36)
      at processImmediate (internal/timers.js:461:21)

```

EDIT: On further investigation, I found that `changeIndexByHash` does not contain actor `0000` in the re-hydrated version of the document. https://github.com/automerge/automerge/blob/main/backend/new.js#L1523

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -1421,5 +1421,32 @@ describe('Automerge', () => {
         diffs: {objectId: '_root', type: 'map', props: {bird: {[`1@${actor}`]: {type: 'value', value: 'Goldfinch'}}}}
       }])
     })
+
+    it.only('should apply changes from another document initialized with the same schema', () => {
+        function getDocument() {
+            let change = Automerge.getLastLocalChange(Automerge.change(Automerge.init('0000'), { time: 0 }, (doc) => {
+                doc.panels = []
+            }))
+            let [doc, ] = Automerge.applyChanges(Automerge.init(), [change])
+            return Automerge.load(Automerge.save(doc))
+        }
+        let s1 = getDocument()
+        let s2 = getDocument()
+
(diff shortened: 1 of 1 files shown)


ept:
Yes, we definitely need to improve this. See also issues #4 and #374 for previous discussions of the same problem. If you want to have a go at fixing this, I can help you get started.

ept:
Thanks for the test case. I think I've fixed the bug on #446 — can you give it a try?

Issue #449, opened 2021-12-01
Binary data format version compatibility


darrenmce:
Hi, firstly thanks for this excellent library - on the whole its been an absolute pleasure to work with.

I seem to have ran into my first (potential) gotcha.

I've been using `preview1` and storing documents in binary data format, however upon upgrading to `preview6` it seems it cannot load documents that have non-integers in them.

a small example:

```json
  "dependencies": {
    "automerge": "1.0.1-preview.6",
    "automerge-preview1": "npm:automerge@1.0.1-preview.1"
  }
```

```js
const AutomergePreview6 = require('automerge');
const AutomergePreview1 = require('automerge-preview1');

const oldDoc = AutomergePreview1.from({ num: -931.75 });
AutomergePreview6.load(AutomergePreview1.save(oldDoc));
```

output:
```
RangeError: Invalid length for floating point number: 4
```

I was hoping someone could help me find the best approach in moving forward. Thanks!

ept:
Hi @darrenmce, thanks for bringing this up. This is the result of dropping support for 32-bit floats in #371, but I forgot that this would make earlier documents unreadable. Sorry for that oversight. I will make you a patch to let you upgrade your document with 32-bit floats to 64-bit floats, which should then be compatible with future versions of Automerge.

darrenmce:
Thanks @ept , sounds great.

darrenmce:
Hi @ept, I just tried it out on a few of my documents and it looks good, was able to load them into `preview6` without issue - much appreciated! 

Thanks again!

ept:
Okay, glad to hear it!

PR #450, opened 2021-12-09
Fix error "RangeError: Invalid length for floating point number: 4"

ept:
This PR is for @darrenmce, who reported a problem in ​#449: if an Automerge document was created with a version between v1.0.1-preview.1 and v1.0.1-preview.3, and the document contained floating-point numbers, then that document would no longer be readable in the latest version of Automerge, but instead throw an exception "RangeError: Invalid length for floating point number: 4". This is due to the fact that those Automerge versions sometimes encoded numbers as 32-bit (single precision floats), and in #371 we moved to using only 64-bit (double) floats.

This patch adds a function `Automerge.upgrade(doc)` which takes an old document containing 32-bit floats, and upgrades it to use 64-bit numbers instead. The document can be given either as a Uint8Array (if it's the result of `Automerge.save()`) or an array of Uint8Arrays (if it's the result of `Automerge.getAllChanges()`).

Note: this PR is not intended to be merged into main, because I don't want to add the `Automerge.upgrade` function to the API permanently, and the upgrade is a one-time operation. Anyone who needs to upgrade their document should check out the code on this PR.

### DIFF
--- a/@types/automerge/index.d.ts
+++ b/@types/automerge/index.d.ts
@@ -67,6 +67,9 @@ declare module 'automerge' {
   function load<T>(data: BinaryDocument, options?: any): Doc<T>
   function save<T>(doc: Doc<T>): BinaryDocument
 
+  function upgrade(changes: BinaryChange[]): BinaryChange[]
+  function upgrade(data: BinaryDocument): BinaryDocument
+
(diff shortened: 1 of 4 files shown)


ept:
I will close this PR as it does not need to be merged, but anybody who needs to upgrade their Automerge docs can find the code to do so here.

Issue #437, opened 2021-10-02
Request of helpfull hints : Automerge Sync Protocol Implementation


raphael10-collab:
In the attempt to re-write https://github.com/cudr/slate-collaborative/blob/master/packages/backend/src/AutomergeBackend.ts
that was working with previous automerge versions,
I'm trying to implement what has been outlined here: https://github.com/automerge/automerge/blob/main/SYNC.md

This is what I've done so far (forgive me for eventual errors):

    interface peer_docId_hash {
      [docId: string]: string
    }
    
    interface syncStatesType {
      [peerId: string]: peer_docId_hash // a hash of [source][docId] containing in-memory sync state
    }
    
    type docsHandledByPeer = string[]
    
    interface peerType {
      peerId: string,
      docs?: docsHandledByPeer,
    }
    
    class AutomergeBackend {
      peers: peerType[] // peers discovered by hyperswarm listening to the same "subject"
    
      syncStates: syncStatesType = {}
    
      backends = {} // a hash by [docId] of current backend values : ??????????
    
      syncPeersListFeed = (peer: peerType): syncStatesType => {
        let sst: syncStatesType = {}
    
        let peersKeysList = Object.keys(this.peers[0])
    
        for (let j = 0; j < this.peers.length; j++) {
          let peer_j= this.peers[j]
          let peer_j_0 = peer_j[peersKeysList[0]]
          sst[peer_j_0] = {}
        }
        return sst
    
      constructor(peers) {
        this.peers = peers
        this.syncStates = this.syncPeersListFeed(this.peers[0])
      }
    
    
      peerPrint (peer: peerType) {
        //console.log(".........................")
    
        let peersKeysList = Object.keys(this.peers[0])
    
        for (let i = 0; i < peersKeysList.length; i++) {
          console.log("peer[peersKeysList[" + i + "]]: ", peer[peersKeysList[i]])
        }
        //Object.keys(peer).forEach(key => {
          //console.log(peer[key])
        //})
      }
    
      random_id = () => { // it mimics hash function
        return Math.random().toString(36).substr(2);
      }
    
      syncPeerDocsFeed = (peer: peerType): void => {
    
        let peersKeysList = Object.keys(this.peers[0])
    
        let peer_docsListLength = peer[peersKeysList[1]].length;
    
        for (let j = 0; j < peer_docsListLength; j++) {
          this.syncStates[peer[peersKeysList[0]]][peer[peersKeysList[1]][j]] = this.random_id()
        }
      }
[...]

PR #458, opened 2022-01-03
Added a simple example to help understanding the sync protocol

FelixSelter:
I added a simple example of two peers connecting and syncing a document.  
This should make things more clear and avoid issues like this one #437  
Previously you had to investigate the tests to understand this correctly  

I hope this helps.
Great work you all

### DIFF
--- a/SYNC.md
+++ b/SYNC.md
@@ -180,3 +180,221 @@ The Automerge sync protocol behaves correctly when synchronizing with multiple p
 
 ### Sequencing Network Communication
 Many communication protocols require that only one of the two peers initiates the exchange. With Automerge this is not the case: messages between peers can interleave arbitrarily, though each peer's messages should be delivered in-order to the recipient.
+  
+
+## A full example how to synchronize **one** document
+
+```js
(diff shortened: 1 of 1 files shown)


ept:
Hi @FelixSelter, thanks for writing this up. I think the code example is not quite correct though. If I read it correctly, you're taking a message from `Automerge.generateSyncMessage()` and broadcasting it to several connected peers. That's not safe, because a sync message is always specific to two particular peers. You need a separate syncState for each connected peer, and generate a separate sync message to send to each connected peer.

FelixSelter:
I will correct that. Strange it actually worked for me with multiple peers

FelixSelter:
@ept Can you please have a look at it again. It still seems wrong. 
1 syncs with 2
2 applies the changes and sees that it does not need to sync back to 1

However if peer1 calls sync again without changing anything it will send out changes to peer2 which also applies them.

timfpark:
Thanks for doing this @FelixSelter - this is super helpful to understand the protocol with implementation details.

PR #463, opened 2022-01-20
explicitly expose Lamport ordering in a test

orionz:
Had a bug in the rust version with lamport ordering being wrong - none of the tests caught it - here are a few that would have caught this sooner.


ept:
Great, thank you @orionz!

PR #471, opened 2022-02-13
remove limitation that merge requires different actors

nikgraf:
When I chatted with @ept he mentioned that the restriction is artificial and was put there to stop people accidentally misusing the API. Unfortunately it is a limitation for certain use-cases.

My use-case: I'm having an end-to-end encrypted protocol where `save` is leveraged to create Snapshots of the current document that are stored remotely. Now if a client has a

- local version
- fetches the latest snapshot to apply it

You can't merge the documents together. I also can imaging architectures that leverage `save` to place document into different storages and need to merge them.

Of course this already is an advanced use-case and could be worked around directly leveraging `applyChanges`. 

Let me know if you think the restriction is important enough or it's fine to remove.

_Note_: `yarn test` is passing

### DIFF
--- a/README.md
+++ b/README.md
@@ -413,8 +413,7 @@ and it is the essence of what Automerge is all about.
 
 `Automerge.merge(doc1, doc2)` is a related function that is useful for testing. It looks for any
 changes that appear in `doc2` but not in `doc1`, and applies them to `doc1`, returning an updated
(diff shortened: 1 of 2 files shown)


ept:
Okay, this looks good to me. Thanks @nikgraf!

PR #476, opened 2022-03-08
Clone issue fix where changesEncoders were not getting copied to cloned object

anshuljhawar:
changesEncoders were not getting copied to the cloned object. This resulted in losing the in-memory information/state of the changesEncoders for the cloned object.


ept:
Hi @anshuljhawar, thanks, this seems like an important fix. However, I would prefer not to add any dependencies on other packages if they can be avoided. Could we do the cloning without lodash?

PR #492, opened 2022-05-24
Declaration specifies more operations than supported

JonhyOliveira:
The declaration files specifies support for ALL operations specified in nodejs' Array declaration file. In fact, not all are actually supported, for instance `array.push(element)` is not. At https://github.com/automerge/automerge/blob/ad5104200c5bb0e6525f30a90d100eee26fa927b/frontend/text.js#L167 we have all the supported operations. (Spoiler alert: push is not there 🙁)

### DIFF
--- a/@types/automerge/index.d.ts
+++ b/@types/automerge/index.d.ts
@@ -86,10 +86,13 @@ declare module 'automerge' {
     deleteAt?(index: number, numDelete?: number): List<T>
   }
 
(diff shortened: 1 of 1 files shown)


JonhyOliveira:
An alternative would be for Text to implement the Array<string> interface and specify an implementation for these unsupported operations (maybe push is an insert at the end?)

PR #531, opened 2023-02-10
allow referencing existing objects when part of the conflict set

mpareja:
Issue https://github.com/automerge/automerge-rs/issues/526 demonstrated that it was not possible to resolve a conflict by choosing one of the conflicting objects or arrays. The following change allows setting a nested object to be an existing object reference so long as the referenced object id is currently listed as one of the conflicts for the key.

### DIFF
--- a/frontend/context.js
+++ b/frontend/context.js
@@ -229,7 +229,14 @@ class Context {
    */
   createNestedObjects(obj, key, value, insert, pred, elemId) {
     if (value[OBJECT_ID]) {
-      throw new RangeError('Cannot create a reference to an existing document object')
+      const object = this.updated[obj] || this.cache[obj]
+      const conflicts = object && object[CONFLICTS][key]
+      const valueInConflicts = conflicts && Object.values(conflicts)
+        .some(v => v && v[OBJECT_ID] === value[OBJECT_ID])
+
+      if (!valueInConflicts) {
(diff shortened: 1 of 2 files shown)


frontend/context.js
```diff
@@ -229,7 +229,14 @@ class Context {
    */
   createNestedObjects(obj, key, value, insert, pred, elemId) {
     if (value[OBJECT_ID]) {
-      throw new RangeError('Cannot create a reference to an existing document object')
+      const object = this.updated[obj] || this.cache[obj]
```
mpareja, line 232: This falls down when the array has nested objects.

A `makeList` operation is added and `createNestedObjects` is called recursively. The `updated` and `cache` maps haven't yet learned about the newly (?) added node.

```js
it('should allow referencing an existing list of objects that is part of the conflict set', () => {
  s2 = Automerge.clone(s1)
  s1 = Automerge.change(s1, doc => doc.list = [{ field: 'a' }])
  s2 = Automerge.change(s2, doc => doc.list = [{ field: 'b' }])

  const result = Automerge.merge(s1, s2)
  const conflicts = Automerge.getConflicts(result, 'list')

  // Should not throw "Cannot create a reference to an existing document object"
  Automerge.change(result, doc => {
    doc.list = Object.values(conflicts)[0]
  })
})
```

